PR #6, opened 2020-04-02
Return empty extensions array when no extension for the given module

thomasdelhomenie:
In the extension registry, when loading extensions of a given module, if the module was not registered, the loading failed.

This fix make sure to not fail in that case and return an empty array.


aelbenney (approved):

ahamdi (approved):
LGTM

PR #10, opened 2020-04-03
PLF-8727: dropup menu according to position

Jihed525:
If the displayed menu would exceed the screen height(The click y position + right click Menu offsetHeight), the added booststrap class "dropup" will assure that this menu will appear in upper direction, if not, this class would be removed in order to restore the original display mode "dropdown"

### DIFF
--- a/web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
+++ b/web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
@@ -251,10 +251,13 @@
 	      break;
 	    default:
 	      // if it isn't fit to be showed down BUT is fit to to be showed up
-	      if ((eXo.core.Mouse.mouseyInClient + ctxMenuContainer.offsetHeight) > browserHeight
-	          && (intTop > ctxMenuContainer.offsetHeight)) {
-	        intTop -= ctxMenuContainer.offsetHeight;
+	      if ((eXo.core.Mouse.mouseyInClient + intTop) > browserHeight )
+          {
+              $(".uiRightClickPopupMenu").addClass("dropup");
 	      }
+		else {
+            $(".uiRightClickPopupMenu").addClass("dropup");
+            	}
 	      break;
 	    }
 	
--- a/web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
+++ b/web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
@@ -256,7 +256,7 @@
               $(".uiRightClickPopupMenu").addClass("dropup");
 	      }
 		else {
-            $(".uiRightClickPopupMenu").addClass("dropup");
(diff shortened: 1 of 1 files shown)


web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
```diff
@@ -251,10 +251,15 @@
 	      break;
 	    default:
 	      // if it isn't fit to be showed down BUT is fit to to be showed up
-	      if ((eXo.core.Mouse.mouseyInClient + ctxMenuContainer.offsetHeight) > browserHeight
-	          && (intTop > ctxMenuContainer.offsetHeight)) {
-	        intTop -= ctxMenuContainer.offsetHeight;
+	      if ((eXo.core.Mouse.mouseyInClient + intTop) > browserHeight )
+	         {
+	        if(document.getElementById('NavigationNodePopupMenu'))document.getElementById('NavigationNodePopupMenu').classList.add("dropup");
+	        if(document.getElementById('ECMContextMenu'))document.getElementById('ECMContextMenu').classList.add("dropup");
```
ahamdi: We can not use a class from ECMS module here, the fix should work without attaching it to the module context.

ahamdi (changes requested):

web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
Jihed525: Well understood, thank you.

Jihed525: > the fix should work without attaching it to the module context.

so even when using this "NavigationNodePopupMenu" isn't allowed, isn't it?

Jihed525: well understood, thank you.

web/eXoResources/src/main/webapp/javascript/eXo/webui/UIRightClickPopupMenu.js
```diff
@@ -251,10 +251,13 @@
 	      break;
 	    default:
 	      // if it isn't fit to be showed down BUT is fit to to be showed up
-	      if ((eXo.core.Mouse.mouseyInClient + ctxMenuContainer.offsetHeight) > browserHeight
-	          && (intTop > ctxMenuContainer.offsetHeight)) {
-	        intTop -= ctxMenuContainer.offsetHeight;
+	      if ((eXo.core.Mouse.mouseyInClient + intTop) > browserHeight )
+          {
+              $(".uiRightClickPopupMenu").addClass("dropup");
```
ahamdi, line 256: Why do you add the same CSS class in both conditions ?

Jihed525: My bad so sorry, it should be removeClass!!!!!!!

ahamdi (approved):

PR #14, opened 2020-04-16
PLF-8765: Oauth integration: sign up/in with linkedin account iss…

lbengayed:
…ue migrate to Oauth2
 ON May 1, 2019, LinkedIn deprecate the use of its v1 APIs.
Applications requesting Version 1.0 APIs experience issues as services begin to remove.
we need to migrate to Version 2.0 of our APIs and OAuth 2.0.


component/web/oauth-common/src/main/java/org/gatein/security/oauth/linkedin/LinkedinProcessorImpl.java
```diff
@@ -46,9 +51,12 @@
     private final String redirectURL;
     private final String apiKey;
     private final String apiSecret;
+    private final String scope="r_liteprofile r_emailaddress w_member_social";
```
ahamdi:
```suggestion
    private final String scope = "r_liteprofile r_emailaddress w_member_social";
```

component/web/oauth-web/src/main/java/org/gatein/security/oauth/web/OAuthProviderFilter.java
```diff
@@ -109,7 +110,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 }
                 interactionState = getOauthProviderProcessor().processOAuthInteraction(httpRequest, httpResponse, scopeToUse);
             }
-        } catch (OAuthException ex) {
+        } catch (OAuthException | ExecutionException | InterruptedException ex) {
             log.warn("Error during OAuth flow with: " + ex.getMessage());
```
ahamdi: Change it to error and print the exception please

pom.xml
```diff
@@ -73,6 +73,8 @@
     <npm.version>6.1.0</npm.version>
 
     <!-- to move to gatein-dep -->
+    <version.scribejava-apis>6.9.0</version.scribejava-apis>
+    <version.scribejava-core>6.9.0</version.scribejava-core>
     <version.scribe>1.3.5</version.scribe>
```
ahamdi: to remove this one if replaced by newer version.

ahamdi (changes requested):
- Please add a unit test
 - Please check that during the build, the right version of scribe library is used

pom.xml
lbengayed: it used to import Token and OAuthService to verify if the token null or no [https://github.com/Meeds-io/gatein-portal/pull/14/files#diff-4b6c1978f7bb2d0fad7bf290fe443d51R114](url)

component/web/oauth-common/src/main/java/org/gatein/security/oauth/linkedin/LinkedinProcessorImpl.java
```diff
@@ -158,8 +157,8 @@ public LinkedinAccessTokenContext validateTokenAndUpdateScopes(LinkedinAccessTok
 
     @Override
     public void saveAccessTokenAttributesToUserProfile(UserProfile userProfile, OAuthCodec codec, LinkedinAccessTokenContext accessToken) {
-        String encodedAccessToken = codec.encodeString(accessToken.accessToken.getToken());
-        String encodedAccessTokenSecret = codec.encodeString(accessToken.accessToken.getSecret());
+        String encodedAccessToken = codec.encodeString(accessToken.accessToken.getAccessToken());
+        String encodedAccessTokenSecret = codec.encodeString(accessToken.accessToken.getAccessToken());
```
ahamdi: What is the difference between those variables ? aren't they the same ?

component/web/oauth-common/src/test/java/org/gatein/security/oauth/test/TestOAuthPrincipalProcessor.java
```diff
@@ -105,4 +116,29 @@ public void testLinkedInGenerateGateInUser() {
         assertEquals("lastName", user.getLastName());
         assertEquals("displayName", user.getDisplayName());
     }
+
+    public void testprocessOAuthInteraction() throws IOException, ExecutionException, InterruptedException {
+
```
ahamdi, line 120: This tests the Scribe Java API, we need to test LinkedinProcessorImpl.java instead

ahamdi (changes requested):

ahamdi (approved):
LGTM

PR #16, opened 2020-04-27
 Allow to configure LDAP/AD integration with exo.properties

thomasdelhomenie:
This improvement allows to configure LDAP/AD integration with exo.properties only, so without an extension and custom PicketLink configurations.
It implements the chapter "Configuration" of the spec https://community.exoplatform.com/portal/g/:spaces:platform_4/exo_platform/wiki/Simplified_LDAP_configuration.

The following parameters are available:
* **exo.plidm.password.class**: FQN of the class used to encode the users password - defaults to "org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder" - Existing parameter
* **exo.plidm.password.hash**: algorithm used to encode the users password - defaults to SHA-256 - Existing parameter
* **exo.plidm.password.secureRandomAlgorithm**: algorithm used to generate random values for password salt - defaults to "SHA1PRNG"
* **exo.idm.search.notCaseSensitive**: allow case insensitive search when searching entities (users and groups) - defaults to "true"
* **exo.ldap.type**: type of LDAP server - "ldap" or "ad" - defaults to "", meaning no LDAP/AD integration
* **exo.ldap.url**: URL to the LDAP server - defaults to "http://localhost:389"
* **exo.ldap.admin.dn**: full DN of the admin user - defaults to "cn=admin"
* **exo.ldap.admin.password**: password of the admin user - defaults to ""
* **exo.ldap.connection.pool**: is connection pooling used ? - defaults to "true"
* **exo.ldap.connection.pool.authentication**: list of space-separated authentication types of connections that may be pooled, alias for com.sun.jndi.ldap.connect.pool.authentication - possible values are "none", "simple", and "DIGEST-MD5" - defaults to "none simple"
* **exo.ldap.connection.pool.maxsize**: maximum number of connections per connection identity that can be maintained concurrently, alias for com.sun.jndi.ldap.connect.pool.maxsize - defaults to 300000
* **exo.ldap.connection.pool.timeout**: number of milliseconds that an idle connection may remain in the pool without being closed and removed from the pool, alias for com.sun.jndi.ldap.connect.pool.timeout - defaults to 0 (no timeout)
* **exo.ldap.connection.pool.protocol**: list of space-separated protocol types of connections that may be pooled, alias for com.sun.jndi.ldap.connect.pool.protocol - possible values are "plain" and/or "ssl" - defaults to "plain ssl"
* **exo.ldap.search.timelimit**: number of milliseconds that a search may last - defaults to 10000
* **exo.ldap.users.base.dn**: semicolon-separated list of full DNs of the objects containing the users (no value means that users are not stored in the LDAP). An empty value means users are not synchronized - by default only 1 DN is defined to "ou=users,dc=company,dc=org".
* **exo.ldap.users.id.attributeName**: attribute used to authentified the users - defaults to "uid" for LDAP and "sAMAccountName" for AD
* **exo.ldap.users.password.attributeName**: attribute used for password for users authentication - defaults to "userPassword" for LDAP and "unicodePwd" for AD
* **exo.ldap.users.filter**: filter used to fetch the users - defaults to "(&(uid={0})(objectClass=User))"
* **exo.ldap.users.attributes.{firstName|lastName|email}.mapping** - mapping of the mandatory users attributes (firstName, lastName and email) - defaults to "cn" for LDAP and "givenName" for AD for firstName, "sn" for lastName and "mail" for email
* **exo.ldap.users.attributes.custom.names**: comma-separated list of custom users attributes (for example "department,city") - defaults to ""
* **exo.ldap.users.attributes.{name}.mapping**: mapping of the custom user attribute - defaults to the name of the custom attribute
* **exo.ldap.users.attributes.{name}.type**: type of the custom user attribute - possible values are "text" or "binary" - defaults to "text"
* **exo.ldap.users.attributes.{name}.isRequired**: is the attribute required - possible values are "true" or "false" - defaults to "false"
* **exo.ldap.users.attributes.{name}.isMultivalued**: is the attribute multi-valued - possible values are "true" or "false" - defaults to "false"
* **exo.ldap.users.search.scope**: scope of the search - possible values are "base", "one", "subtree" - defaults to "subtree"
* **exo.ldap.groups.base.dn**: semicolon-separated list of full DNs of the objects containing the groups (no value means that groups are not stored in the LDAP). An empty value means groups are not synchronized - by default only 1 DN is defined to "ou=groups,dc=company,dc=org".
* **exo.ldap.groups.id**: attribute used to authentified the groups - defaults to "cn"
* **exo.ldap.groups.filter**: filter used to fetch the groups - defaults to "(&(cn={0})(objectClass=Group))"
* **exo.ldap.groups.attributes.custom.names**: comma-separated list of custom groups attributes (for example "description,city") - defaults to ""
* **exo.ldap.groups.attributes.{name}.mapping**: mapping of the custom group attribute - defaults to the name of the custom attribute
* **exo.ldap.groups.attributes.{name}.type**: type of the custom group attibute - possible values are "text" or "binary" - defaults to "text"
* **exo.ldap.groups.attributes.{name}.isRequired**: is the attribute required - possible values are "true" or "false" - defaults to "false"
* **exo.ldap.groups.attributes.{name}.isMultivalued**: is the attribute multi-valued - possible values are "true" or "false" - defaults to "false"
* **exo.ldap.groups.search.scope**: scope of the search - possible values are "base", "one", "subtree" - defaults to "subtree"
* **exo.ldap.groups.parentMembershipAttributeName**: LDAP attribute that defines children of IdentityObject. This will be used to retrieved relationships from IdentityObject entry. Option is required if IdentityObjectType can be part of relationship. - defaults to "member"
* **exo.ldap.groups.isParentMembershipAttributeDN**: defines if values of attribute defined in parentMembershipAttributeName are fully qualified LDAP DNs - defaults to "true"
* **exo.ldap.groups.childMembershipAttributeName**: LDAP attribute that defines parents of IdentityObject. This will be used to retrieved relationships from IdentityObject entry. Good example of such attribute in LDAP schema is 'memberOf' - defaults to ""
* **exo.ldap.groups.childMembershipAttributeDN**: defines if values of attribute defined in childMembershipAttributeName are fully qualified LDAP DNs - defaults to "false"
* **exo.ldap.groups.rootGroup**: root group to bind LDAP/AD groups, all LDAP/AD groups will be available under this group. It must end with "/*". Root group ("/*") cannot be used - defaults to "/platform/*"

Technical choices:
* the org.exoplatform.services.organization.idm.PicketLinkIDMService component will now contain one value-param by supported directory type, named "config.{type}" where {type} is the of directory type (possible values are "ldap", "ad" and ?)
* the default config (no ldap) is kept in the existing "config" value-param". This allows to keep backward compatibility, even if the customers change the value of this config parameter
* most of the parameters are simple substitutes thanks to ${} notation
* since PicketLink configuration does not allow to define optional attributes by configuration, the choosen solution has been to dynamically update the PicketLink configuration at startup, based on properties (ExoIdentityStoreConfigurationContext)
* `sAMAccountName` is used as the default authentication id field for AD. Two fields are generally used: `sAMAccountName` or `userPrincipalName`. According to http://www.kouti.com/tables/userattributes.htm it should be `userPrincipalName` for recent Windows Server versions, but after a test, `sAMAccountName` contains the username whereas `userPrincipalName` contains the email, so `sAMAccountName` is used in order to login with the username by default.
* all the examples and ldif files have been deleted since they are obsolete. Documentation must be updated accordingly.

TODO:

- [x] unit tests
- [ ] allow to map groups to the root (/)
- [x] make root group configurable
- [x] implement multiple ctxDNs as a comma-separated list instead of finite list of properties ? The solution based on comma-separated value has not been used because ctxDNS value is not only used in the methods searchIdentityObjects, and we could  probably use the same approach than for custom attributes (update conf dynamically at startup).

QUESTIONS:
* for AD, should we use `unicodePwd` or `userPassword` for passwrd field ? According to https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/3983110d-4d24-4a59-baeb-db5b863a92c6 they are similar.


rdenarie:
Ok for me ATM. The only thing I can see is the parametization for nested groups (properties parentMembershipAttributeName and isParentMembershipAttributeDN)

In fact, there are 2 properties which are opposite to the 2 previous :
childMembershipAttributeName
isChildMembershipAttributeDN

I have difficulties to find docs about that.

Theses properties are used in very few cases. So I'm not sure we should add this now. WDYT ?

thomasdelhomenie:
Yes I agree, we should parameterize `parentMembershipAttributeName` and `isParentMembershipAttributeDN`.

> I have difficulties to find docs about that.

Hell yes! :unamused: 

I found this in the [source code](https://github.com/picketlink/picketlink-idm/blob/1.4/picketlink-idm-docs/ReferenceGuide/en/ReferenceGuide.xml):

* **parentMembershipAttributeName**: LDAP attribute that defines children of IdentityObject. This will be used to retrieved relationships from IdentityObject entry. Option is required if IdentityObjectType can be part of relationship. 
* **isParentMembershipAttributeDN**: defines if values of attribute defined in parentMembershipAttributeName are fully qualified LDAP DNs.
* **childMembershipAttributeName**: LDAP attribute that defines parents of IdentityObject. This will be used to retrieved relationships from IdentityObject entry. Good example of such attribute in LDAP schema is 'memberOf'
* **childMembershipAttributeDN**: defines if values of attribute defined in childMembershipAttributeName are fully qualified LDAP DNs.

I will parameterize all of them, it does not cost anything.

component/identity/src/main/java/org/exoplatform/services/organization/idm/PicketLinkIDMServiceImpl.java
```diff
@@ -123,6 +140,7 @@ public PicketLinkIDMServiceImpl(ExoContainerContext exoContainerContext, InitPar
 
             this.configMD = JAXB2IdentityConfiguration.createConfigurationMetaData(confManager
                     .getInputStream(this.config));
+            //this.configMD.getIdentityStores().get(0).getSupportedIdentityTypes().get(0).getOption("ctxDN")
```
ahamdi: Could you remove this line ?

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
```diff
@@ -163,12 +177,13 @@
...
-                  <string>/</string>
-                </key>
-                <value>
-                  <string>root_type</string>
-                </value>
+                <key><string>/</string></key>
+                <value><string>root_type</string></value>
+              </entry>
+
+              <entry>
```
ahamdi, line 163: Is there a way to make this configured from external too ? I mean enabling group configuration or not ? and deciding to which group they will be mapped ?

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
```diff
@@ -100,6 +104,16 @@
...
       </value-param>
 
+      <value-param>
+        <name>config.ldap</name>
+        <value>war:/conf/organization/picketlink-idm/picketlink-idm-ldap-config.xml</value>
+      </value-param>
+
+      <value-param>
+        <name>config.ad</name>
+        <value>war:/conf/organization/picketlink-idm/examples/picketlink-idm-msad-readonly-config.xml</value>
```
ahamdi: Could you move this to the same path as picketlink-idm-ldap-config.xml ? it will be easier to check which files are used .

ahamdi (changes requested):

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
thomasdelhomenie: Yes, I did not implement AD conf, it is in the TODO list. I will put files in the same folder.

component/identity/src/main/java/org/exoplatform/services/organization/idm/PicketLinkIDMServiceImpl.java
thomasdelhomenie: Sure.

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
thomasdelhomenie: I think so, I will check that. By default I want to map the groups at the root level (see TODO list), but I did not succeed so far.

ahamdi: OK

component/identity/src/main/java/org/exoplatform/services/organization/idm/PicketLinkIDMServiceImpl.java
thomasdelhomenie: Done.

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
thomasdelhomenie: Done.

ahamdi (dismissed):
LGTM

rdenarie (dismissed):
LGTM

ahamdi (approved):

PR #17, opened 2020-04-28
Update license header for Meeds

uocnb:


boubaker (changes requested):
@hbenali please avoid merging this PR since it has a lot of files. The PR is unreadable and should be split to multiple ones.

PR #19, opened 2020-05-01
Exclude transitive dependency of scribejava

ahamdi:
After adding library scribejava 6.9.0, the build of social project was failing. It was due to having a new different version of com.fasterxml.jackson.core:jackson-databind dependency that was added as a transitive dependency of com.github.scribejava


boubaker (dismissed):

boubaker (approved):
After checking the additional libraries imported in Tomcat/lib, I have found that only the following third party libraries are added:
* scribejava-apis-6.9.0.jar
* scribejava-core-6.9.0.jar

to replace this on:
* scribe-1.3.5.jar

So it's ok for me, no additional third-party libraries were added (They were already existant in Tomcat/lib of eXo Server)

PR #28, opened 2020-06-10
Fix user language saving and move configuration

boubaker:
(Sorry the PR seems as huge while it's just formatting)
The method 'saveLocaleToUserProfile' wasn't triggered all time when the user switches to a different language. This modification ensure to compare last used locale of user with current locale defined for HTTP request and save it if changed switch `LocalePolicy` service implementation.


ghost (approved):

ahamdi (approved):
LGTM

PR #34, opened 2020-06-24
Optimize DB datasource usage for quartz

boubaker:
Quartz is using Database as default store even in standalone mode. This mode is useful only for Cluster to be able to synchronize jobs execution through nodes. For standalone mode, it's useless and consumes a lot of resources for nothing. This modification will use Quartz Jobs management in-memory in standalone and DB for cluster mode only.


ahamdi (approved):
LGTM
This needs to be documented IMO

PR #35, opened 2020-06-24
Deactive cache usage during setEnable function

rdenarie:
If we don't deactivate cache, in user event listeners like preEnable and postEnable, the cache could have a false value


mdiha (approved):
LGTM

PR #52, opened 2020-07-13
MAINT-26699: Alert message still displayed

AzmiTouil:
The alert message should be automatically removed after 5 seconds


ahamdi (commented):
Could you check if the error messages are correct ?

AzmiTouil:
> Could you check if the error messages are correct ?

yes ,correct

ahamdi (approved):
LGTM

PR #57, opened 2020-07-21
MAINT-26909: Add personalNameValidation for first and last name

AzmiTouil:
Only letters, spaces, "-" or "'" are allowed for first and last name


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -33,15 +33,15 @@
 
   public static final String             WRONG_USER_PASSWORD_ERROR_CODE = "WRONG_USER_PASSWORD";
 
-  public static final UserFieldValidator USERNAME_VALIDATOR             = new UserFieldValidator("usermame", true);
+  public static final UserFieldValidator USERNAME_VALIDATOR             = new UserFieldValidator("usermame", true,false);
```
boubaker: please format the added block and add UT

boubaker (approved):

PR #59, opened 2020-07-22
Maint-26952: check validity of role name and description length

hassenelkamel-exo:
- add a check to insure that the name must be between "3" and "30" characters, The length of the text in field "Role Description" must be between "3" and "255" characters.

### DIFF
--- a/component/portal/src/test/java/org/exoplatform/portal/rest/MembershipTypeRestResourcesTest.java
+++ b/component/portal/src/test/java/org/exoplatform/portal/rest/MembershipTypeRestResourcesTest.java
@@ -21,6 +21,8 @@ public class MembershipTypeRestResourcesTest extends BaseRestServicesTestCase {
 
   private static final String MEMBERSHIP_TYPE_2 = "mt2";
 
+  private static final String MEMBERSHIP_TYPE_3 = "mt3";
+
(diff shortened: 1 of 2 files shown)


component/portal/src/test/java/org/exoplatform/portal/rest/MembershipTypeRestResourcesTest.java
```diff
@@ -70,6 +72,12 @@ public void testCreateMembershipType() throws Exception {
     data.put("name", "NEW_MEMBERSHIP_TYPE");
     response = getResponse("POST", "/v1/membershipTypes", data.toString());
     assertNotNull(response);
+    assertEquals(400, response.getStatus());
```
ahamdi, line 83: Please cover error cases in the UT

ahamdi (changes requested):

component/portal/src/test/java/org/exoplatform/portal/rest/MembershipTypeRestResourcesTest.java
hassenelkamel-exo: done

ahamdi (approved):

PR #64, opened 2020-07-24
MAINT-26699: Alert message still displayed

AzmiTouil:
To fix the problem, we should store "**this**" outside the timeout  function.


ahamdi (approved):
LGTM

PR #66, opened 2020-08-10
39908 : Add Open Id SSO system

rdenarie:


component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdAccessTokenContext.java
```diff
@@ -0,0 +1,66 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
ahamdi: Shouldn't we change the header of the class ?

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdPrincipalProcessor.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) 2012 eXo Platform SAS.
```
ahamdi:
```suggestion
 * Copyright (C) 2021 eXo Platform SAS.
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdPrincipalProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public User convertToGateInUser(OAuthPrincipal principal) {
+        String email = principal.getEmail();
+        String username = principal.getUserName();
+        if(email != null) {
+            int index = email.indexOf('@');
+            if(index > 0) {
+                username = email.substring(0, index);
+            }
+        }
```
ahamdi, line 38: I think we do not need to change th eusername unless it exists or it is not available in the OAuthPrincipal
If the username is not found, we may use the firstname + "." + lastname and avoid using the email username, WDYT ?

web/portal/src/main/webapp/WEB-INF/conf/sso/oauth-configuration.xml
```diff
@@ -129,6 +129,50 @@
...
     </init-params>
   </component>
+  
+  <component>
+    <key>org.gatein.security.oauth.openid.OpenIdProcessor</key>
+    <type>org.gatein.security.oauth.openid.OpenIdProcessorImpl</type>
+    <init-params>
+      <value-param>
+        <name>clientId</name>
+        <value>${gatein.oauth.openid.clientId}</value>
```
ahamdi: IMO we need to use the pattern exo.oauth.openid instead of gatein.oauth.openid

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -0,0 +1,481 @@
...
+            OAuth2AccessToken token = new OAuth2AccessToken(decodedAccessToken, decodedAccessTokenSecret);
+            return new OpenIdAccessTokenContext(token);
+        }
+    }
+    
+    @Override
+    public void removeAccessTokenFromUserProfile(UserProfile userProfile) {
+        OAuthPersistenceUtils.removeLongAttribute(userProfile, OAuthConstants.PROFILE_LINKEDIN_ACCESS_TOKEN, false);
+        OAuthPersistenceUtils.removeLongAttribute(userProfile, OAuthConstants.PROFILE_LINKEDIN_ACCESS_TOKEN_SECRET, false);
+    }
```
ahamdi: IMO PROFILE_OPEN_ID instead of PROFILE_LINKEDIN

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -0,0 +1,481 @@
...
+        OAuthPersistenceUtils.saveLongAttribute(encodedAccessToken, userProfile, OAuthConstants.PROFILE_OPEN_ID_ACCESS_TOKEN,
+                                                false, chunkLength);
+        
+    }
+    
+    @Override
+    public OpenIdAccessTokenContext getAccessTokenFromUserProfile(UserProfile userProfile, OAuthCodec codec) {
+        String encodedAccessToken = OAuthPersistenceUtils.getLongAttribute(userProfile,
+                                                                           OAuthConstants.PROFILE_OPEN_ID_ACCESS_TOKEN, false);
+        String encodedAccessTokenSecret = OAuthPersistenceUtils.getLongAttribute(userProfile, OAuthConstants.PROFILE_LINKEDIN_ACCESS_TOKEN_SECRET, false);
```
ahamdi: IMO PROFILE_OPEN_ID instead of PROFILE_LINKEDIN

ahamdi (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdAccessTokenContext.java
rdenarie: I remove it. 
I dont know exactly which header I should put

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdPrincipalProcessor.java
rdenarie: In fact, this is call when the user successfully logs in against his identityProvider, but not exists in exo.
In this case, which username should we use ? 
I decide to take the first part of the email. 
In fact, this login will be never use by the user as he will always logs with his email with the IdentityProvider.

web/portal/src/main/webapp/WEB-INF/conf/sso/oauth-configuration.xml
rdenarie: Done. In addition in documentation I put exo.

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
rdenarie: done

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
rdenarie: done

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdPrincipalProcessor.java
ahamdi: An OpenID authenticated user could login with his username & password if he wishes. He needs just to reset his password, but he won't know what username was assigned to him. That is why I thought that his first+lastname could be a good option WDYT ?

ahamdi (approved):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdPrincipalProcessor.java
rdenarie: Yes, he can, but once he have used and oauth system to first login and create his account, dont you think he will still use the oauth mecanism ? In addition, if he really want to login directly in exo, I think he will use his email, has he can now.

PR #67, opened 2020-09-04
MAINT-28042 : Implement function to retrieve all membershipTypes for a group

ahamdi:
This is an implementation for retrieving all membershipTypes linked to a specified group.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/MembershipDAOImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/MembershipDAOImpl.java
@@ -716,7 +716,7 @@ public class MembershipDAOImpl extends AbstractDAOImpl implements MembershipHand
     }
 
     @Override
-    public List<MembershipType> findMembershipTypesByGroup (String groupId) throws Exception{
+    public List<MembershipType> findMembershipTypesByGroup(String groupId) throws Exception{
         if (log.isTraceEnabled()) {
             Tools.logMethodIn(log, LogLevel.TRACE, "findMembershipTypesByGroup", new Object[] { "group", groupId});
         }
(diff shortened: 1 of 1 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/MembershipDAOImpl.java
```diff
@@ -714,6 +715,53 @@ public Membership findMembership(String id) throws Exception {
         return null;
     }
 
+    @Override
+    public List<MembershipType> findMembershipTypesByGroup (String groupId) throws Exception{
```
azayati:
```suggestion
    public List<MembershipType> findMembershipTypesByGroup(String groupId) throws Exception{
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/MembershipDAOImpl.java
```diff
@@ -714,6 +715,53 @@ public Membership findMembership(String id) throws Exception {
...
+        } catch (Exception e) {
+            handleException("Identity operation error: ", e);
+        }
+
+        if (log.isTraceEnabled()) {
+            Tools.logMethodOut(log, LogLevel.TRACE, "findMembershipTypesByGroup", roleTypes);
+        }
+
+        List<MembershipType> membershipTypes = new ArrayList<>();
+        if (roleTypes != null && !roleTypes.isEmpty()) {
```
azayati, line 750: IMO CollectionUtils.isEmpty can be used instead

azayati (changes requested):
See review

component/identity/src/main/java/org/exoplatform/services/organization/idm/MembershipDAOImpl.java
ahamdi: The same code will be used for both cases, I prefer avoiding that to depend less on utilities classes as soon as they are not required.

azayati (approved):
LGTM

PR #74, opened 2020-10-01
Fix/improve files management when binary not available

rdenarie:


component/file-storage/src/main/java/org/exoplatform/commons/file/resource/FileSystemResourceProvider.java
```diff
@@ -132,6 +132,12 @@ public InputStream getStream(String name) {
     try {
       return new FileInputStream(getFile(name));
     } catch (IOException e) {
+      try {
+        log.error("Unable to read binary content from disk for file "+getFile(name).getPath() +". The binary content is not "
```
azayati:
```suggestion
        log.error("Unable to read binary content from disk for file " + getFile(name).getPath() + ". The binary content is not "
```

component/file-storage/src/main/java/org/exoplatform/commons/file/resource/FileSystemResourceProvider.java
```diff
@@ -132,6 +132,12 @@ public InputStream getStream(String name) {
...
+                      + "accessible, it was removed, or may have been quarantined by an antivirus.",e);
+      } catch (IOException ioException) {
+        log.error("Unable to get file object for name "+name,ioException);
```
azayati:
```suggestion
        log.error("Unable to get file object for name " + name, ioException);
```

component/portal/src/main/java/org/exoplatform/portal/branding/BrandingServiceImpl.java
```diff
@@ -264,7 +264,7 @@ public Logo getLogo() {
 
   @Override
   public Logo getDefaultLogo() {
-    if (this.logo == null) {
+    if (this.logo == null || this.logo.getData()==null) {
```
azayati:
```suggestion
    if (this.logo == null || this.logo.getData() == null) {
```

azayati (changes requested):
Format

azayati (approved):
LGTM

PR #77, opened 2020-10-06
Add kernel profiles and css class to pages containers

boubaker:
When uninstalling some addons, the pages that defines portlets / applications defined by deleted addons gets buggy with error message 'Portlet not found'. Knowing that each addon defines its own Kernel profile when installed, this improvement makes it easy to disable a container with its applications in rendering phase, when the Kernel profile isn't present.
Another attribute is added in Container definition in pages.xml file that allows to define a list of CSS Classes to use in container Parent.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -824,7 +824,7 @@
               </property>
               <property>
                 <name>exo.profiles</name>
-                <value>${surefire.exo.profiles}</value>
+                <value>${surefire.exo.profiles},test</value>
               </property>
             </systemProperties>
           </configuration>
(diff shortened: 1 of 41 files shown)


pom.xml
```diff
@@ -824,7 +824,7 @@
               </property>
               <property>
                 <name>exo.profiles</name>
-                <value>${surefire.exo.profiles}</value>
+                <value>${surefire.exo.profiles},test</value>
```
rdenarie, line 827: Are you to put the value test ?

boubaker: This is just an Exo Kernel profile that is added to be used after in Surefire Unit Tests

azayati (approved):
LGTM

rdenarie (approved):

ahamdi (approved):
LGTM

PR #80, opened 2020-10-08
Fix refreshing WebUI Page cache after saving page on storage

boubaker:
When saving page on storage through backend services or REST API, the WebUI components are still cached and aren't refreshed. This fix listens to events triggered from storage layer to update WebUI Component when a modification is made on page


ahamdi (dismissed):

ahamdi (approved):
LGTM again

PR #82, opened 2020-10-13
Switch default Cluster configuration to use TCP instead of UDP

boubaker:
Since UDP is deprecated for production environments, we will have to switch default configuration to use TCP by default instead.


bourasom (approved):
LGTM

PR #89, opened 2020-10-30
Optimize fetched resources caching in browser side

boubaker:
This PR aims to improve resources fetching by :
- Allowing to reuse RememberMe Cookie on REST calls when user session timed out (by changing parent path of cookie)
- Adding a lastUpdateTime on Branding CSS to allow cache it efficiently in browser
- Group fetching of `baseGRP` resources that englobes multiple base functions that are widely used in all pages
- Fix Service worker to retrieve the same headers of retrieved HTML documents from Server End


ahamdi (approved):
LGTM

PR #90, opened 2020-10-31
TASK-35860 Improve front end performances

boubaker:
This PR defines a Service worker to manage efficiently resources caching and prepare to PWA feature for making Portals accessible in offline mode (not fully supports offline mode yet, but a preparation of it by using Service Worker caching).

1. In defined Service Worker (that used [workbox](https://developers.google.com/web/tools/workbox))
* we cache efficiently the static resources coming from server independently from returned HTTP Headers. for development mode, the JS and CSS files aren't handled by service worker. 
* we manages the DOM parts (portlets or part of portlets) loading optimization, by allowing to render the last version of a chunk of html at the same time than rendering the page. In fact when the page is loaded, before returning the HTML to browser in order to display it, the Service Worker will search for custom HTML tag `<v-cacheable-dom-app cache-id="CACHE_ID"></v-cacheable-dom-app>` and search whether a cache key `CACHE_ID` exists in browser cache or not, if it exists, it will be replaced by the cached DOM, else, the tag will be ignored by the browser anyway. After displaying the Application, the application itself will have to put DOM inside browser cache to use it next loading time.

2. In addition to the evolution of Skeleton loader of applications, an optimization has been made on UITabContainer.gtmpl file that uses Vuetify to render tabs. In fact, to improve UX while Vue application is not loaded yet, this fix will add some Vuetiy classes to avoid glitches while the page is loading.

3. This PR allows to reuse RememberMe Cookie on REST calls when user session timed out (by changing parent path of cookie)
4. This PR allows to use a lastUpdateTime on Branding CSS REST endpoint to allow cache it efficiently in browser side using eTag.


web/portal/src/main/webapp/WEB-INF/conf/pwa/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
```
ahamdi: We need to remove the manifest IMO

web/portal/src/main/webapp/WEB-INF/conf/pwa/pwa-configuration.xml
```diff
@@ -0,0 +1,37 @@
...
+    <key>PWAConfigurationProperties</key>
+    <type>org.exoplatform.container.ExtendedPropertyConfigurator</type>
+    <init-params>
+      <properties-param>
+        <name>PWAConfigurationProperties</name>
+        <description>PWA configuration properties</description>
+        <property name="pwa.service.worker.enabled" value="${pwa.service.worker.enabled:true}" />
+        <property name="pwa.service.worker.path" value="${pwa.service.worker.path:war:/conf/pwa/service-worker.js}" />
+        <property name="pwa.manifest.enabled" value="${pwa.manifest.enabled:true}" />
+        <property name="pwa.manifest.path" value="${pwa.manifest.path:war:/conf/pwa/manifest.json}" />
```
ahamdi: If Manifest will be disabled, we could remove those configurations too

ahamdi (changes requested):

web/portal/src/main/webapp/WEB-INF/conf/pwa/manifest.json
boubaker: Done, thanks @ahamdi

web/portal/src/main/webapp/WEB-INF/conf/pwa/pwa-configuration.xml
boubaker: Done, thanks @ahamdi

ahamdi (approved):

PR #109, opened 2020-11-09
MAINT-36199 : 'No file selected' is displayed in change folder's icon image

Rdidmohamedelhadi:


ahamdi (changes requested):
You can not remove an element created by another template. we need tgo do it inside using UIThumbnailForm itself

PR #111, opened 2020-11-12
Task-36396: On the onboarding screen, display the password condition defined by the property

rdenarie:
Before this fix, the passwordCondition displayed was the one in localization file
But we can define a message by configuration, this message should be displayed here


azayati (approved):
LGTM until QA validation

PR #113, opened 2020-11-13
Task-36001: Allow login with username or email

azayati:


rdenarie:
@azayati Did you check if it works with and LDAP or AD ?

rdenarie (dismissed):

rdenarie:
Did you remove the webui properties because it will be treaten by crowdin ?

component/web/security/src/main/java/org/exoplatform/web/login/LoginServlet.java
```diff
@@ -171,6 +176,24 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
         int status;
         if (req.getRemoteUser() == null) {
             if (username != null && password != null) {
+                // email authentication
+                if (username.contains("@")) {
```
plamarque, line 180: and what if username contains an @ ?

azayati: basically, the username with @ is forbidden, but even if the username contains @ (with dirty creation), we will not find a user with email == username (with the @) and the login will then be executed normally

rdenarie (approved):

PR #119, opened 2020-11-18
Task-36579 : Add type on token

rdenarie:
Before this commit, the token is used for forgot-password, onboarding, and external users registration. So the token for external users registration can be used to make the onboarding
This commit add a type on token so that the forgot-password token can be used only for forgot password, and so on for other types


azayati (approved):
LGTM

PR #125, opened 2020-11-23
Task-36931: When forgetPassword, resetPassword, onboarding or external registration css changes, users have to refresh cache

rdenarie:
Before this fix, the plf version number was not present in css url
This fix add the assets version in css url


azayati (approved):
LGTM

PR #131, opened 2020-11-26
Task-37127 : add the possibility to paginate LDAP search result

rdenarie:
In synchronization, we need to get all LDAP users. If there are more user than the default limit, we should ne able to use pagination on results.
Currently, we cannot activate pagedResult.
This commit add the possibility to configure that point.


boubaker (dismissed):

azayati (approved):
LGTM

PR #136, opened 2020-12-03
Task-37266: Change rules about error message for incorrectness with login and forgot password.

ahmadmido:


azayati (approved):
LGTM after QA validation

PR #138, opened 2020-12-04
Task-37081: Don't allow login with an email associated to many users

azayati:
This will avoid login with email associated to many users, an error message is displayed to ask login with username instead


AzmiTouil (approved):
LGTM

PR #139, opened 2020-12-07
Task-37383: Display error message when we try to invite external user with same username of an already deleted user

azayati:
Currently when we try to invite external user with same first name and last name of an already deleted user, a blank page is displayed with an error caused by wrong label used. This fix will display the right error message


ahmadmido (dismissed):

AzmiTouil (approved):
LGTM

PR #141, opened 2020-12-09
TASK-37899 Enhance extension registry to trigger event when extensions updated

boubaker:
When updating extensionRegistry asynchronously from different modules, the other modules should be triggered for this update to allow retrieving newly registered extensions. This way, the extension points in all portlets will can be aware, when listening to adequate event, of any change in extensionRegistry.


ahamdi (approved):
LGTM

PR #149, opened 2021-01-07
Add log when changing password

rdenarie:
We have no log when user change password
This commit add logs

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/recovery/DefaultChangePasswordConnector.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/recovery/DefaultChangePasswordConnector.java
@@ -9,8 +9,8 @@ import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
 import javax.naming.directory.ModificationItem;
 
-import org.gatein.common.logging.Logger;
-import org.gatein.common.logging.LoggerFactory;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
 import org.gatein.portal.idm.impl.repository.ExoFallbackIdentityStoreRepository;
 import org.picketlink.idm.api.IdentitySession;
 import org.picketlink.idm.impl.api.session.IdentitySessionImpl;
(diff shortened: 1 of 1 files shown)


component/web/security/src/main/java/org/exoplatform/web/login/recovery/DefaultChangePasswordConnector.java
```diff
@@ -126,6 +129,9 @@ private void changeExternalPassword(User user, String password) throws Exception
       }
       mods[0] = new ModificationItem(DirContext.REPLACE_ATTRIBUTE, mod0);
       ctx.modifyAttributes(userDN, mods);
+      long totalTime = System.currentTimeMillis() - startTime;
+      log.info("service={} operation={} parameters=\"user:{}\" status=ok duration_ms={}",
```
boubaker, line 133: we don't need this, in fact we shouldn't, even, we must not (RGPD).
If you want to add stats for this, you can use exo-analytics

azayati (approved):
LGTM

component/web/security/src/main/java/org/exoplatform/web/login/recovery/DefaultChangePasswordConnector.java
rdenarie: In fact, we have not access to identitymanager in portal, so we cannot display something else.
I'm checking with @plamarque about RGPD compliance for logs

rdenarie: As we discuss, we need to display a log, not collect a stat. 
I try to find a way to pseudonymize it, by for example using identityId. But IdentityManger is in social, so I dont have acces to it. 
So if you have an idea, I take it. Else, I think I will let as is.

plamarque: It's fine to log username for GDPR as long as we keep only the necessary time. It means as long as logs are rotated, we should be lawful.

boubaker: I 'm not sure that this is the best way to add a trace to users behaviors. Using Production log is usually for troubleshooting problems and this should be in a debug level in order to help diagnosing. IMO, using analytics as an audit system is what should be done to trace Users interactions with the systems and their actions instead of using logs.

PR #150, opened 2021-01-07
fix/TASK-25736: Unbind on click event before assigning the new one

aelbenney:
Every node in the navigation menu is bound twice the `onClick` event. When you click on copy, paste or any action, it will be fired twice.
This fix will make sure that every menu action has only one listener for the click event.


ahamdi:
@aelbenney Could you add a description ?

ahamdi (approved):
LGTM

PR #162, opened 2021-02-06
40139 : Load home node of current site when adding a page

ahamdi:
When adding a new page from a page included in global site, the new page will be added to the Global site.
This PR makes sure to select the current site if the navigation belongs to the Global site

### DIFF
--- a/build-config/pom.xml
+++ b/build-config/pom.xml
@@ -27,13 +27,13 @@
   <parent>
     <groupId>org.exoplatform.gatein.portal</groupId>
     <artifactId>exo.portal.parent</artifactId>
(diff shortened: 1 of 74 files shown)


boubaker:
I think this was already fixed by this commit : https://github.com/Meeds-io/gatein-portal/blob/80ca793b2ddb03586258ff1163949f0267a4edab/webui/portal/src/main/java/org/exoplatform/portal/url/PortalURLContext.java (I wasn't aware of this task in Maintenance Board)
Have you noticed another strange behavior ?

ahamdi:
@boubaker yes, the other fix will get the right URI after page is created. This PR will make sure the page is c reated under the current site instead of ading it to Global site.

PR #166, opened 2021-02-11
 38509 : Fix configuration to solve groups upgrade issue fc856f0 

ahamdi:
The identity type of all groups under /platform were mistakenly changed from .platform to GROUP with configuration added to enable configuring LDAP using exo.properties file
In thi sPR we add some properties to make it possible to define values for all IDM identity types, to help old installations use previous values which makes it possible to retrieve correctly users by groups.
The new properties are : 

- exo.group.platform.identity.type : it defines the identity type of groups under /platform group.Before meeds 1.0 , the value was '.platform' , starting meeds 1.0 the value is GROUP. Default value : GROUP
- exo.ldap.groups.rootGroup is the group under which all LDAP/AD groups will be imported. Default value was changed to /ldap-groups/* instead of /platform/*
- exo.ldap.groups.identity.type : it defines the identity type for groups imported from LDAP/AD. Default value is LDAP_MAPPED_GROUP

### DIFF
--- a/web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
+++ b/web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
@@ -160,9 +160,17 @@
                 <key><string>/</string></key>
                 <value><string>root_type</string></value>
               </entry>
+              <!-- exo.group.platform.identity.type will define the identity type of groups under /platform group.
+                Before meeds 1.0 , the value was '.platform' , starting meeds 1.0 the value is GROUP -->
               <entry>
-                <key><string>${exo.ldap.groups.rootGroup:/platform/*}</string></key>
-                <value><string>GROUP</string></value>
+                <key><string>/platform/*</string></key>
+                <value><string>${exo.group.platform.identity.type:GROUP}</string></value>
+              </entry>
+              <!-- exo.ldap.groups.rootGroup is the group under which all LDAP/AD groups will be imported
+                   exo.ldap.groups.identity.type is the identity type for groups imported from LDAP/AD-->
+              <entry>
+                <key><string>${exo.ldap.groups.rootGroup:/ldap-groups/*}</string></key>
+                <value><string>${exo.ldap.groups.identity.type:LDAP_MAPPED_GROUP}</string></value>
               </entry>
             </map>
           </field>
--- a/build-config/pom.xml
+++ b/build-config/pom.xml
@@ -27,13 +27,13 @@
   <parent>
     <groupId>org.exoplatform.gatein.portal</groupId>
     <artifactId>exo.portal.parent</artifactId>
-    <version>6.2.x-SNAPSHOT</version>
+    <version>6.2.x-maintenance-SNAPSHOT</version>
   </parent>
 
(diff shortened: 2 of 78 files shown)


web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
```diff
@@ -160,9 +160,17 @@
                 <key><string>/</string></key>
                 <value><string>root_type</string></value>
               </entry>
+              <!-- exo.group.platform.identity.type will define the identity type of groups under /platform group.
+                Before meeds 1.0 , the value was '.platform' , starting meeds 1.0 the value is GROUP -->
               <entry>
-                <key><string>${exo.ldap.groups.rootGroup:/platform/*}</string></key>
-                <value><string>GROUP</string></value>
+                <key><string>/platform/*</string></key>
+                <value><string>${exo.group.platform.identity.type:GROUP}</string></value>
```
boubaker, line 167: I understand that this commit had made simplification of LDAP installation https://github.com/Meeds-io/gatein-portal/pull/16, but it introduced a change to root group of LDAP groups. In other terms, it will add all LDAP mapped groups under /platform. By making this change, the LDAP mapped groups will be moved to `/ldap-groups/`. Will this induce a modification in the structure of groups of existing installations using LDAP mapping ?

I think, the changes made in this PR and the PR https://github.com/Meeds-io/gatein-portal/pull/16 needs a detailed **Upgrade** instructions of different use cases. (@hbenali @bourasom FYI )

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
```diff
@@ -160,9 +160,17 @@
...
+              </entry>
+              <!-- exo.ldap.groups.rootGroup is the group under which all LDAP/AD groups will be imported
+                   exo.ldap.groups.identity.type is the identity type for groups imported from LDAP/AD-->
+              <entry>
+                <key><string>${exo.ldap.groups.rootGroup:/ldap-groups/*}</string></key>
```
boubaker, line 172: Shouldn't we initialize the group `/ldap-groups` in `org.exoplatform.services.organization.OrganizationDatabaseInitializer` ? What's the impact when not doing this ?

ahamdi: The group will be created automatically when we start the synchronization.

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
ahamdi: Currently, all groups under /platform/* are of identity type GROUP. Thus with this change I wanted to keep this changed beahvior for customers using eXo platform 6.0+ 
For old customers, they will be able to restore the default value (.platform) and their users will be imported under their /platform/* groups.
If a customer has a group mapping, he will be able to define the group under which he will import all his groups. I put the default value to /ldap-groups/ and didn't want to use an existing group

boubaker (approved):

PR #167, opened 2021-02-13
Disable debug mode for Service worker

boubaker:
Debug mode for Service worker is not needed event in development mode. It makes the browser console verbose.


ahamdi (dismissed):
Yes indeed

PR #170, opened 2021-02-15
 38509 : Fix LDAP configuration to solve groups upgrade issue  (#166)

mzorai:
The identity type of all groups under /platform were mistakenly changed from .platform to GROUP with configuration added to enable configuring LDAP using exo.properties file
In thi sPR we add some properties to make it possible to define values for all IDM identity types, to help old installations use previous values which makes it possible to retrieve correctly users by groups.
The new properties are :

    exo.group.platform.identity.type : it defines the identity type of groups under /platform group.Before meeds 1.0 , the value was '.platform' , starting meeds 1.0 the value is GROUP. Default value : GROUP
    exo.ldap.groups.rootGroup is the group under which all LDAP/AD groups will be imported. Default value was changed to /ldap-groups/* instead of /platform/*
    exo.ldap.groups.identity.type : it defines the identity type for groups imported from LDAP/AD. Default value is LDAP_MAPPED_GROUP

### DIFF
--- a/web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
+++ b/web/portal/src/main/webapp/WEB-INF/conf/organization/idm-configuration.xml
@@ -160,9 +160,17 @@
                 <key><string>/</string></key>
                 <value><string>root_type</string></value>
               </entry>
+              <!-- exo.group.platform.identity.type will define the identity type of groups under /platform group.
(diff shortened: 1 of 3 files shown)


hbenali (approved):

mzorai:
@ahamdi could you please check the build failure here?

PR #173, opened 2021-02-18
Task-40830 : When using initial account wizard, the user password is …

rdenarie:
…exposed in a GET request

Before this fix, when we use the initial account wizard, we make a GET request on /portal/login to automatically log the user. The password is added as query String of this request and so exposed in access log.
This fix moves the AccountSetup servlet from commons to portal, and logs the user directly after form submit without any additionnal request, so that the password is not exposed.


ahmadmido (approved):
LGTM

PR #177, opened 2021-03-08
Task-41341 : Change userManegement page style.

ahmadmido:
This PR will change the style of usersManegement : 
- Separate actions delete - edit - role.
- Change the icon on the role column.
- The delete icons are available only for superusers and are red.
- Restore the display of the user's status: Reuse the switch buttons for this column and remove the switch button enable/disable from the edit user drawer.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
@@ -492,7 +492,7 @@ public class UserRestResourcesV1 implements ResourceContainer {
   }
 
   @GET
-  @Path("checkIsSuperUser")
+  @Path("isSuperUser")
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed("administrators")
   @ApiOperation(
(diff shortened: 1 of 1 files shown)


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -491,6 +491,69 @@ public Response getUserMemberships(
     return Response.ok(new CollectionEntity<>(membershipEntities, offset, limit, totalSize)).build();
   }
 
+  @GET
+  @Path("checkIsSuperUser")
```
```suggestion
  @Path("isSuperUser")
```

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -491,6 +491,69 @@ public Response getUserMemberships(
...
+      response = Response.class
+  )
+  @ApiResponses(
+      value = {
+          @ApiResponse(code = 200, message = "Request fulfilled"),
+          @ApiResponse(code = 400, message = "Invalid query input"),
+          @ApiResponse(code = 500, message = "Internal server error due to data encoding"),
+      }
+  )
+  public Response updateUserStatus(@Context HttpServletRequest request, UserRestEntity userEntity) throws Exception {
```
rdenarie: We already have a function updateUser in this rest service. Is it possible to used it instead of creating a new one ?

rdenarie (changes requested):

rdenarie (approved):

PR #178, opened 2021-03-10
Task-41306 : Add source column on UserManegement

ahmadmido:
This PR will add a new column on UserManegmenet table -> Source.
This column will check if a user is an internal user or a synchronized user, if is internal we display an icon with eXo logo and if's synchronized LDAP/AD we display an icon with a tag SYNC.


rdenarie (changes requested):
Please check comment on social request. We want to display the last sync date, not the creation date, which are not the same

PR #183, opened 2021-03-25
TASK-41619 : Synchronisation AD - User status

GouadriaHanen:
added enabled to attributes while retrieving users from AD

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
@@ -100,11 +100,11 @@ public class IDMUserListAccess implements ListAccess<User>, Serializable {
         if (fullResults == null) {
             users = listQuery(index, length);
 
-            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED) && !isDBOnly) {
+            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED)) {
                 users = filterUserByStatus(users, this.userStatus, index, length);
             }
         } else {
-            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED) && !isDBOnly) {
+            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED)) {
                 //Need to check all returned users is enabled
                 users = filterUserByStatus(fullResults, this.userStatus, index, length);
             } else{
@@ -179,7 +179,7 @@ public class IDMUserListAccess implements ListAccess<User>, Serializable {
               } else {
                 fullResults = listQuery(0, 0);
 
-                if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED) && !isDBOnly) {
+                if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED)) {
                   result = filterUserByStatus(fullResults, this.userStatus, 0, fullResults.size()).size();
                 } else {
                   result = fullResults.size();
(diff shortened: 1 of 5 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
```diff
@@ -100,11 +100,11 @@ public IDMUserListAccess(UserQueryBuilder userQueryBuilder, int pageSize, boolea
         if (fullResults == null) {
             users = listQuery(index, length);
 
-            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED) && !isDBOnly) {
+            if ((this.userStatus == UserStatus.ENABLED || this.userStatus == UserStatus.DISABLED)) {
```
ahamdi: Why did you remove the boolean isDbOnly ?

component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
```diff
@@ -222,7 +222,7 @@ PicketLinkIDMOrganizationServiceImpl getOrganizationService() {
             while (iterator.hasNext() && result.size() < length) {
               org.picketlink.idm.api.User user = iterator.next();
                 Attribute attr = getIDMService().getIdentitySession().getAttributesManager().getAttribute(user.getKey(), EntityMapperUtils.USER_ENABLED);
-                if ((userStatus == UserStatus.ENABLED && attr == null) || (userStatus == UserStatus.DISABLED && attr != null && attr.getValue().toString().equals("false"))) {
+                if ((userStatus == UserStatus.ENABLED && attr == null) || (userStatus == UserStatus.DISABLED && attr != null && attr.getValue().toString().equals("false"))|| (userStatus == UserStatus.ENABLED && attr != null && attr.getValue().toString().equals("true"))) {
```
ahamdi: Could you combine this check with the first one in the same instruction ?

component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
```diff
@@ -793,7 +793,7 @@ exclude USER type (search for all group types) */
                   //Nothing
                }
                /** Begin eXo customization : PLF-7270**/
-               if (entry.getValue() == null || entry.getValue().length == 0) {
+               if ((entry.getValue() == null || entry.getValue().length == 0) && mappedAttributeName!= "enabled") {
```
ahamdi: Could you tell more about this check ?

component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
```diff
@@ -3256,54 +3256,39 @@ private void filterByAttributesValues(Collection<IdentityObject> objects, Map<St
                //Nothing
             }
 
-            // If the attribute key is enable, consider its absence as it was equals to true
```
ahamdi: Could youi tell more about this change ?

ahamdi (changes requested):

component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
GouadriaHanen: removed to always filter by status when retrieving by status

component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
GouadriaHanen: "enabled" is added to attributes so the query becomes " and not exists(from io.attributes as attrTableJoinName where attrTableJoinName.name = :enabled " will return an empty slice

component/identity/src/main/java/org/exoplatform/services/organization/idm/IDMUserListAccess.java
GouadriaHanen: No, it is not the same

component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
GouadriaHanen: an internal user may have its status null means that he is enabled

ahamdi (approved):

PR #188, opened 2021-04-02
Task-42659 : Fix the bug of exo user not allowed to change password when property exo.portal.allow.change.external.password not displayed.

ahmadmido:
The problem is due to the property: exo.portal.allow.change.external.password is null in isSynchronizedUserAllowedToChangePassword, so I add a check to verify if the property is null or not.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
@@ -528,7 +528,7 @@ public class UserRestResourcesV1 implements ResourceContainer {
     if (user == null) {
       throw new WebApplicationException(Response.Status.BAD_REQUEST);
     }
-    boolean allowChangePassword = Boolean.valueOf(System.getProperty("exo.portal.allow.change.external.password").toString()) || user.isInternalStore();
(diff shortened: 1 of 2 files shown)


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -528,7 +528,7 @@ public Response isSynchronizedUserAllowedToChangePassword(@Context UriInfo uriIn
     if (user == null) {
       throw new WebApplicationException(Response.Status.BAD_REQUEST);
     }
-    boolean allowChangePassword = Boolean.valueOf(System.getProperty("exo.portal.allow.change.external.password").toString()) || user.isInternalStore();
+    boolean allowChangePassword = (System.getProperty("exo.portal.allow.change.external.password") != null && Boolean.valueOf(System.getProperty("exo.portal.allow.change.external.password").toString())) || user.isInternalStore();
```
```suggestion
    boolean allowChangePassword = user.isInternalStore() || (System.getProperty("exo.portal.allow.change.external.password") != null && Boolean.valueOf(System.getProperty("exo.portal.allow.change.external.password").toString()));
```

In fact, the case we will see in majority is the user in internal store, so it is preferable to check it first

rdenarie (changes requested):
Can you also add a unit test

rdenarie (approved):

PR #190, opened 2021-04-13
Task-42613: Advanced filters - connected / not connected users

AzmiTouil:
This will ensure to update the user's profile with the last login time when the user logged in.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
@@ -71,7 +71,7 @@ public class UpdateLoginTimeListener extends Listener<ConversationRegistry, Conv
       }
 
       user.setLastLoginTime(Calendar.getInstance().getTime());
-      userHandler.saveUser(user, false);
+      userHandler.saveUser(user, true);
     } catch (Exception e) {
       LOG.error("Error while updating the last login time for user {}", userId, e);
     } finally {
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
@@ -24,6 +24,7 @@ import org.exoplatform.container.component.RequestLifeCycle;
 import org.exoplatform.services.listener.Asynchronous;
 import org.exoplatform.services.listener.Event;
 import org.exoplatform.services.listener.Listener;
+import org.exoplatform.services.listener.ListenerService;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.services.organization.OrganizationService;
@@ -41,6 +42,8 @@ public class UpdateLoginTimeListener extends Listener<ConversationRegistry, Conv
   private PortalContainer     container;
 
   private OrganizationService organizationService;
+  
+  private ListenerService listenerService;
 
   public UpdateLoginTimeListener(PortalContainer container) {
     this.container = container;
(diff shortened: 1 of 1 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
```diff
@@ -71,7 +71,7 @@ public void onEvent(Event<ConversationRegistry, ConversationState> event) {
       }
 
       user.setLastLoginTime(Calendar.getInstance().getTime());
-      userHandler.saveUser(user, false);
+      userHandler.saveUser(user, true);
```
ahamdi: Why is this change ? this should be a silent update with no events fired !

ahamdi (changes requested):

component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
AzmiTouil: To add the lastLoginTime in the profile properties, so I can filter users by connected / not connected.

ahamdi: Why do you need to enable broadcasting the event ? I do not see a direct relation between the Save event on User entity and the profile update.

rdenarie (dismissed):
Can you add a description to your PR/commit ?

component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
AzmiTouil: yes there is a relation
https://github.com/Meeds-io/gatein-portal/blob/develop/component/identity/src/main/java/org/exoplatform/services/organization/idm/UserDAOImpl.java#L158

https://github.com/Meeds-io/social/blob/develop/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialUserEventListenerImpl.java#L77

ahamdi: I know that relation. I don't get it why we should relaunch all user Listeners to just save a Date property inside the user profile.
Why don't you just load the user profile and add that property in the same time as adding it to the user entity ?

rdenarie: Here we are in gatein, and not haev access to userProfile, which is in social.
I think it is why @AzmiTouil use broadcast.
@AzmiTouil can you confirm ?

AzmiTouil: Yes I confirm

azayati (dismissed):
LGTM

component/identity/src/main/java/org/exoplatform/services/organization/idm/UpdateLoginTimeListener.java
ahamdi: Then I believe we need to add a new listener to catch the Login event and add the property to the user profile on Social instead of broadcasting the User Update event which is too generic.

rdenarie: I'm OK with @ahamdi 's proposition. Let's go to created a new listener in social

rdenarie (changes requested):
I think there is a misunderstood here. 
You just create a new event, fired by an exiting listener UpdateLoginTimeListener.
But this listener is launched on event exo.core.security.ConversationRegistry.register
Why not use the same event ? In your new listener, you just pick the value of user.getLastLoginTime to put it in social profile.

There is no need to create a new event.

AzmiTouil:
> I think there is a misunderstood here.
> You just create a new event, fired by an exiting listener UpdateLoginTimeListener.
> But this listener is launched on event exo.core.security.ConversationRegistry.register
> Why not use the same event ? In your new listener, you just pick the value of user.getLastLoginTime to put it in social profile.
> 
> There is no need to create a new event.

OK

PR #191, opened 2021-04-18
MAINT-35522 : Fix validating terms and condition several times

hakermi:
This PR is fixing the requesting of validating the terms and condition several times:

**ISSUE** : the existing of duplicated stored attribute  "acceptedTermsAndConditions" leads to sending the incorrect accepted version when validating the terms.
**PROPOSED SOLUTION** : Add a check to handle the case of exiting duplicated attributes of "acceptedTermsAndConditions" and sending the max value of them (means : the first updated attribute when executing 
 : _organizationService.getUserProfileHandler().saveUserProfile(userProfile, false)_ )

### DIFF
--- a/component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
+++ b/component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
@@ -120,6 +120,8 @@ public class PatchedHibernateIdentityStoreImpl implements IdentityStore, Seriali
 
    public static final String ALL_GROUPS_TYPE = "@@ALL_GROUPS@@";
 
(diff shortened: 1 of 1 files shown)


component/identity/src/main/java/org/picketlink/idm/impl/store/hibernate/PatchedHibernateIdentityStoreImpl.java
```diff
@@ -120,6 +120,8 @@
 
    public static final String ALL_GROUPS_TYPE = "@@ALL_GROUPS@@";
 
+   private static final String storedTcAttribute = "acceptedTermsAndConditions";
```
boubaker: please use constant variable name pattern

hakermi: Done

PR #192, opened 2021-04-20
Task-43073: Implement a rest method to check if the current user is delegated administrator or not

AzmiTouil:


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -510,6 +510,26 @@ public Response isSuperUser() {
...
+          response = Response.class
+  )
+  @ApiResponses(
+          value = {
+                  @ApiResponse(code = 200, message = "Request fulfilled"),
+                  @ApiResponse(code = 500, message = "Internal server error due to data encoding"),
+          }
+  )
+  public Response isDelegatedAdministrator() {
+    boolean isDelegatedAdministrator = userACL.isUserInGroup("/platform/delegated") && !userACL.isUserInGroup("/platform/administrators");
```
azayati: is it possible to put "/platform/delegated" and "/platform/administrators" as constants ?

azayati (changes requested):
See review

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
AzmiTouil: yes, done !

azayati (approved):
LGTM after QA validation

rdenarie (approved):

PR #195, opened 2021-04-27
Task-43074: users management page for delegated administrators

AzmiTouil:
This PR will Implement a method that searches for users in a given groups according to a search criteria.
This method will be used in user management by the delegated user to get only the users that are members or having any role in the groups for which he is admin.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/UserDAOImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/UserDAOImpl.java
@@ -581,6 +581,65 @@ public class UserDAOImpl extends AbstractDAOImpl implements UserHandler {
     return findUsersByGroupId(groupId, UserStatus.ENABLED);
   }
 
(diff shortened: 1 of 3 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/UserDAOImpl.java
```diff
@@ -26,6 +26,7 @@
 
 import org.gatein.common.logging.LogLevel;
 import org.picketlink.idm.api.*;
+import org.picketlink.idm.api.Group;
```
rdenarie: not needed

rdenarie (changes requested):
Please improve commit message, format should be

"Use a clear commit message of format TASK-XYZ (replace XYZ by task number) and a clear description that explains:
1/ What was the previous behavior (Prior ti this change....)
2/ What is changed/impacted by the current change
3/ (Optional) what other alternative solutions/propositions was studied"****

rdenarie (approved):

rdenarie (approved):

PR #198, opened 2021-04-30
TASK-43352 : fix displaying wrong pages

GouadriaHanen:
A wrong page is displayed when the user is a member of Administrators, Content management, or dlp group : always leads to the default path 
fixed removed displaying the default Path


ahamdi (changes requested):
This is an expected behavior of the current version of Meeds/eXo platform and is used to redirect automatically user requests to their first accessible pages when the requested page is missing or inaccessible.

Removing the auto-redirection will change a current behavior, we need to study and provide a better solution

PR #211, opened 2021-05-17
TASK-26265 Fix generating sources archive

boubaker:
Delete the flag `<useReleaseProfile>false</useReleaseProfile>` that should be used to diable generating sources of project when releasing. This flag will be used by meeds project to generate a bundle of meeds projects sources.


azayati (approved):
LGTM

PR #212, opened 2021-05-20
Task-44021 : Problem on MFA redirection

rdenarie:
Before this fix, in inPageNavigation, when the fetch see a 302 redirect, it follows automatically the redirect, but the url is not update and stay on the intermediate url (the one before the redirect)

This fix moves the instruction window.history.replaceState after the redirect, and set the current url to the last redirect result


web/eXoResources/src/main/webapp/javascript/eXo/portal/InPageNavigation.js
```diff
@@ -107,6 +106,7 @@
           })
             .then(resp => {
               if (resp && resp.status == 200) {
+                window.history.replaceState('', resp.url, resp.url);
```
boubaker: Prior to this change, when the response code is different from 200, the page will be reloaded with new URL.
By using this change we may have a problem which is when having a response code different from 200, the current same page will be reloaded instead of the new one.

rdenarie: What I can propose is : 
Configure the fetch to not follow the 302 redirect, and if we are in case of 302, then reload the page. 
WDYT @boubaker ?

rdenarie: Updated

boubaker (approved):

PR #214, opened 2021-05-21
TASK-44182: fixed redirection security problem in login page URL

nidhalnidhal:
i modified the request handler in login servlet by filtering any suspicious uri    so that the user cannot be redirected to any uri outside of the current context path

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/LoginServlet.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/LoginServlet.java
@@ -287,7 +287,7 @@ public class LoginServlet extends AbstractHttpServlet {
 
         try {
             URI uri = new URI(initialURI);
-            if (uri.isAbsolute() && !(uri.getHost().equals(req.getServerName()))) {
+            if ((uri.getHost() != null) && !(uri.getHost().equals(req.getServerName()))) {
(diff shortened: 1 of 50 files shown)


component/web/security/src/main/java/org/exoplatform/web/login/LoginServlet.java
```diff
@@ -287,7 +287,8 @@ public String getContextPath() {
 
         try {
             URI uri = new URI(initialURI);
-            if (uri.isAbsolute() && !(uri.getHost().equals(req.getServerName()))) {
+            // sanitizing url from suspicious uri
+            if (!uri.getPath().equals(req.getContextPath())) {
```
ahamdi: We can check just for the server name in both cases : 
```suggestion
            if (!(uri.getHost().equals(req.getServerName()))) {
```

ahamdi (changes requested):

component/web/security/src/main/java/org/exoplatform/web/login/LoginServlet.java
nidhalnidhal: i suggest this change :             if ((uri.getHost() != null) && !(uri.getHost().equals(req.getServerName()))) 
because uri.getHost() is returning null for the case of the standard login page URL . wich is causing a null pointer exception in the evalution.

ahamdi (approved):

PR #218, opened 2021-05-28
Maint-44780:fix changing branding logo

Jihed525:
Used the same upload method already used in ecms module to upload files

### DIFF
--- a/cdi/contexts/pom.xml
+++ b/cdi/contexts/pom.xml
@@ -26,7 +26,7 @@
   <parent>
     <groupId>org.exoplatform.gatein.portal</groupId>
     <artifactId>exo.portal.cdi.parent</artifactId>
-    <version>6.2.x-SNAPSHOT</version>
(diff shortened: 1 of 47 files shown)


portlet/exoadmin/package.json
```diff
@@ -35,5 +35,8 @@
     "webpack-cli": "^3.1.1",
     "webpack-dev-server": "^3.11.2",
     "webpack-merge": "^4.1.3"
+  },
+  "dependencies": {
+    "axios": "^0.19.0"
```
ahamdi: We should not use axios. We have to use Javascript Fetch API

ahamdi (changes requested):

portlet/exoadmin/package.json
Jihed525: understood

hbenali (approved):

ahamdi (approved):

PR #222, opened 2021-06-01
Disable InPage navigation feature by default

boubaker:
Disable experimental InPage navigation feature by default


ahamdi (approved):
LGTM

PR #224, opened 2021-06-02
TASK-44539 : improved disabled user detection from AD

nidhalnidhal:
Before this fix , disabled user from AD are detected only if the value of UserAccessControl attribute is equal to 514 , so i generalized the detection for all values and now the rule is:  an AD  user is disabled if UserAccessContral attribute is equal to ( (2 to the (power) n )+ 2) ) . this modification took place in EntityMapperUtils class.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/EntityMapperUtils.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/EntityMapperUtils.java
@@ -127,6 +127,8 @@ public class EntityMapperUtils {
 
       // used when populating User from the platform ; it return true or false
       // if it is from AD it always returns false since it is a number .
+      /* in AD , the (2) value represents a disabled user and  Many UserAccessControl values represent an enabled user like 65536 which represent an enabled user
+      whose password never expires : this user is detected as disabled if he has 65538 as UserAccessControl value   */
       Boolean enabled = Boolean.parseBoolean(attrs.get(USER_ENABLED).getValue().toString()) ;
 
       if (isInteger(status))
@@ -146,8 +148,8 @@ public class EntityMapperUtils {
     return changed;
   }
   private static Boolean isUserEnabled (int status) {
-    String BinaryStatus = Integer.toBinaryString(status);
-    if(BinaryStatus.charAt(BinaryStatus.length()-2) == '1') {
+    String binaryStatus = Integer.toBinaryString(status);
+    if(binaryStatus.charAt(binaryStatus.length()-2) == '1') {
       return false;
     }
     return true;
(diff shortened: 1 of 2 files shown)


ahamdi (changes requested):
We need to add a Unit test to cover this class with different samples of status values.

component/identity/src/main/java/org/exoplatform/services/organization/idm/EntityMapperUtils.java
```diff
@@ -145,5 +145,20 @@ public static boolean populateUser(User user, Map<String, Attribute> attrs, bool
     }
     return changed;
   }
-
+  private static Boolean isUserEnabled (int status) {
+    String BinaryStatus = Integer.toBinaryString(status);
```
plamarque:
```suggestion
    String binaryStatus = Integer.toBinaryString(status);
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/EntityMapperUtils.java
```diff
@@ -145,5 +145,20 @@ public static boolean populateUser(User user, Map<String, Attribute> attrs, bool
...
+    if(BinaryStatus.charAt(BinaryStatus.length()-2) == '1') {
```
plamarque:
```suggestion
    if(binaryStatus.charAt(BinaryStatus.length()-2) == '1') {
```

plamarque (commented):
Some explanation of why these codes all represent a disabled user would be helpful too
(leave in code + add unit test should let a reader understand the unobvious)

component/identity/src/test/java/org/exoplatform/services/organization/TestLDAPOrganization.java
```diff
@@ -136,6 +133,37 @@ public void testFindUser() throws Exception {
     assertNull(test2);
     end();
   }
+  public void testEnabledUserDetection() throws Exception{
```
hbenali, line 137:
```suggestion

  public void testEnabledUserDetection() throws Exception{
```
Please format your code. Please check http://developer.exoplatform.org/#id-intellij-settings

component/identity/src/main/java/org/exoplatform/services/organization/idm/EntityMapperUtils.java
```diff
@@ -145,5 +148,20 @@ public static boolean populateUser(User user, Map<String, Attribute> attrs, bool
     }
     return changed;
   }
-
+  private static Boolean isUserEnabled (int status) {
```
hbenali, line 152: Also here please

hbenali (changes requested):
Please format your code

ahamdi (approved):

hbenali (approved):

PR #230, opened 2021-06-07
45419: Replace usage of hibernate hbm2ddl by liquibase script initialization

boubaker:


boubaker:
@ahamdi I let you consider this in maintenance scope and provide a task identifier. Thanks.

ahamdi (approved):

PR #244, opened 2021-06-28
Task-45972 : Add QrCode js library for MFA OTP

rdenarie:
This commit add a shared js library to generate qrcode which will be used by other part of the product


boubaker (dismissed):

web/eXoResources/src/main/webapp/WEB-INF/gatein-resources.xml
```diff
@@ -934,4 +934,12 @@
       <module>jquery</module>
     </depends>
   </module>
+  
+  <module>
+    <name>qrcode</name>
+    <script>
+      <minify>false</minify>
+      <path>/js/qrious.min.js</path>
```
boubaker:
```suggestion
      <path>/javascript/qrious.min.js</path>
```

plamarque (dismissed):
NOGO qrious license is not compatible with meeds : https://github.com/neocotic/qrious/blob/master/LICENSE.md

We need LGPL, MIT, Apache2 or BSD style license. GPL is not compatible with Meeds' LGPL.

rdenarie:
Change to a library with MIT licence :
https://www.npmjs.com/package/qrjs2

ahamdi (approved):

PR #258, opened 2021-08-04
Task-46504 : Mfa Protected Navigations

rdenarie:
Change to configuration restriction by navigation instead of pages

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
@@ -90,7 +90,7 @@ public class NavigationRest implements ResourceContainer {
                                                            .filter(userNavigation -> !userNavigation.getKey()
                                                                                                     .getName()
                                                                                                     .equals("global"))
-                                                           .map(userNavigation -> new ResultUserNavigation(userNavigation))
+                                                           .map(ResultUserNavigation::new)
                                                            .collect(Collectors.toList());
       return Response.ok(allNavs).build();
 
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
@@ -1,28 +1,52 @@
 package org.exoplatform.portal.mop.rest;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
```diff
@@ -180,10 +214,7 @@ private Response getNavigations(HttpServletRequest request,
 
   private static UserNodeFilterConfig getUserFilterConfig(Visibility[] visibilities) {
     UserNodeFilterConfig.Builder builder = UserNodeFilterConfig.builder();
-    builder.withReadWriteCheck()
-           .withVisibility(visibilities)
-           .withTemporalCheck()
-           .withReadCheck();
+    builder.withReadWriteCheck().withVisibility(visibilities).withTemporalCheck().withReadCheck();
```
boubaker: It's better to avoid changing code formatting for not changed parts

component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
```diff
@@ -40,46 +64,58 @@ public NavigationRest(UserPortalConfigService portalConfigService) {
...
+
+    try {
+      UserPortalConfig userPortalConfig = portalConfigService.getUserPortalConfig(portalName, username, userPortalContext);
+      List<ResultUserNavigation> allNavs = userPortalConfig.getUserPortal()
+                                                           .getNavigations()
+                                                           .stream()
+                                                           .filter(userNavigation -> !userNavigation.getKey()
+                                                                                                    .getName()
+                                                                                                    .equals("global"))
+                                                           .map(userNavigation -> new ResultUserNavigation(userNavigation))
```
boubaker: This can really be a very big object, especially for someone who belongs to a lot of spaces, isn't it ?
We may need a pagination here.

component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
rdenarie: Removed.

component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
rdenarie: Yes it is possible.
This improvment will be tracked and done in another task

ahmadmido (approved):

PR #273, opened 2021-08-30
TASK-40995:External user flag is still displayed next to the user fullname

GouadriaHanen:
when adding a user to 'External group ' an external flag must be displayed next to his name in the chat room and chat drawer: the External Flag isn't always displayed because in the actual behavior  setting user property is set when the user login  and not updated when removed from Group,fixed by adding a listener from MembershipEventListener to update the user enabled property one added or removed from 'External group'


rdenarie (changes requested):
Hello
Can you explain this change ? I'm not sure to understand why it is necessary.

GouadriaHanen:
> Hello
> Can you explain this change ? I'm not sure to understand why it is necessary.

it is not necessary it was a misunderstanding! PR to be closed!

rdenarie:
> > Hello
> > Can you explain this change ? I'm not sure to understand why it is necessary.
> 
> it is not necessary it was a misunderstanding! PR to be closed!

So I let you close it :)
Thanks

PR #279, opened 2021-09-14
Task 48683: Disable Service worker by default

NaderSai:


boubaker (dismissed):

boubaker (commented):
Need to be tested on Maintenance FB first isn't it ?

NaderSai:
> Need to be tested on Maintenance FB first isn't it ?

okay

AzmiTouil (approved):

PR #287, opened 2021-10-29
MAINT-49963: Fix the ability of xss attacks on forgotpassord page

hakermi:
**ISSUE**: The initialURL param could be an javascript function injects a dangerous code in the page
**FIX**: Add a check to allow only the right value for this param which is "/portal/login"


AzmiTouil (approved):
LGTM

ahamdi (approved):

PR #290, opened 2021-11-03
Task-50856 : [Active directory sycnhronization] AD group users aren't…

rdenarie:
… synchronized with the eXo group

Before this fix, when an AD group is member of more than one group, he is synchronized at root group level. When a parent of this group is removed (he now have only one parent), he is sync ad children of his parent. But the  information is not propagated to idm database. So, when a user, member of a group like this connect, he is ejected from this group.
This fix allow to detect that a group moves from the root level to a non root level, and propagate the information to the IDM database

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
@@ -202,7 +202,7 @@ public class GroupDAOImpl extends AbstractDAOImpl implements GroupHandler {
             throw new Exception("Group " + jbidParentOriginGroup + " does not exist");
         }
         if (jbidGroupToMove == null) {
-            throw new Exception("Group " + jbidGroupToMove + " does not exists");
+            throw new Exception("Group " + jbidGroupToMove + " does not exist");
         }
 
 
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
@@ -152,6 +152,7 @@ public class GroupDAOImpl extends AbstractDAOImpl implements GroupHandler {
 
     }
 
+    @Override
     public void moveGroup(Group parentOriginGroup, Group parentTargetGroup,Group groupToMove) throws Exception {
         //find ParentOriginGroup
         if (log.isTraceEnabled()) {
@@ -166,7 +167,6 @@ public class GroupDAOImpl extends AbstractDAOImpl implements GroupHandler {
                                                   .findGroup(plParentOriginGroupName,
                                                              orgService.getConfiguration().getGroupType(parentOriginGroup.getParentId()));
         } catch (Exception e) {
-            // TODO:
             handleException("Identity operation error: ", e);
         }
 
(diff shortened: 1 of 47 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -152,6 +152,96 @@ public void addChild(Group parent, Group child, boolean broadcast) throws Except
...
+        try {
+            jbidGroupToMove = getIdentitySession().getPersistenceManager()
+                                                        .findGroup(plGroupToMoveName,
+                                                                   orgService.getConfiguration().getGroupType(groupToMove.getParentId()));
+        } catch (Exception e) {
+            // TODO:
+            handleException("Identity operation error: ", e);
+        }
+
+        //if one is miossing => error
```
ahamdi:
```suggestion
        //if one is missing => error
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -152,6 +152,96 @@ public void addChild(Group parent, Group child, boolean broadcast) throws Except
...
+        if (jbidParentOriginGroup == null) {
+            throw new Exception("Group " + jbidParentOriginGroup + " does not exists");
```
ahamdi:
```suggestion
            throw new Exception("Group " + jbidParentOriginGroup + " does not exist");
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -152,6 +152,96 @@ public void addChild(Group parent, Group child, boolean broadcast) throws Except
...
+        }
+        if (jbidParentTargetGroup == null) {
+            throw new Exception("Group " + jbidParentOriginGroup + " does not exists");
```
ahamdi:
```suggestion
            throw new Exception("Group " + jbidParentOriginGroup + " does not exist");
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -152,6 +152,96 @@ public void addChild(Group parent, Group child, boolean broadcast) throws Except
...
+        }
+        if (jbidGroupToMove == null) {
+            throw new Exception("Group " + jbidGroupToMove + " does not exists");
```
ahamdi:
```suggestion
            throw new Exception("Group " + jbidGroupToMove + " does not exist");
```

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -152,6 +152,96 @@ public void addChild(Group parent, Group child, boolean broadcast) throws Except
...
+//            jbidOriginGroup = getIdentitySession().getPersistenceManager().findGroup(plParentOriginGroupName,
+//                                                                                     orgService.getConfiguration().getGroupType(originGroup.getParentId()));
+//        } catch (Exception e) {
+//            // TODO:
+//            handleException("Identity operation error: ", e);
+//        }
+//
+//        if (jbidOriginGroup == null) {
+//
+//        }
```
ahamdi: COuld you remove this block if it is not needed

rdenarie: done

rdenarie:
cc @ahamdi  can you validate to merge

ahamdi (approved):

PR #291, opened 2021-11-12
Task-42855: Open onlyoffice editor from document explorer is not stable

AzmiTouil:
Prior this change, there is a check if the requested node Path is equal to last non ajax request node Path and is not equal to the last node Path without taking into account the Portlet Parameters, and the server performs a 302 redirect on the last node Path.
Fix : Remove this check if the Portlet Parameters is not empty

### DIFF
--- a/webui/portal/src/main/java/org/exoplatform/portal/webui/workspace/UIPortalApplication.java
+++ b/webui/portal/src/main/java/org/exoplatform/portal/webui/workspace/UIPortalApplication.java
@@ -812,21 +812,21 @@ public class UIPortalApplication extends UIApplication {
     @Override
     public void processAction(WebuiRequestContext context) throws Exception {
         PortalRequestContext pcontext = (PortalRequestContext) context;
-        // String requestURI = pcontext.getRequestURI();
         RequestNavigationData requestNavData = pcontext.getNavigationData();
 
         boolean isAjax = pcontext.useAjax();
 
         if (!isAjax) {
-            if (isAjaxInLastRequest) {
-                isAjaxInLastRequest = false;
-                if (requestNavData.equals(lastNonAjaxRequestNavData) && !requestNavData.equals(lastRequestNavData)) {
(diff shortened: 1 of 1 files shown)


ahamdi:
Why is it different ? why do you test on portlet parameters ? It is not obvious

AzmiTouil:
> Why is it different ? why do you test on portlet parameters ? It is not obvious

The requestNavData does not contain the portlet params, it only contains siteType, siteName, path. In our case the path will be "**oeditor**".
**Exp**: LastNonAjaxRequest =**/portal/dw/oeditor?docId=10c3a2ce7f000101178c1b1e13ccb050&source=peview**
                     CurrentRequest = **/portal/dw/oeditor?docId=10c3da4f7f000101289a5177814af6da&source=peview**
=> Condition valid and the server will redirect on the last node Path **/documents**

webui/portal/src/main/java/org/exoplatform/portal/webui/workspace/UIPortalApplication.java
```diff
@@ -812,21 +812,21 @@ public void processDecode(WebuiRequestContext context) throws Exception {
...
-                isAjaxInLastRequest = false;
-                if (requestNavData.equals(lastNonAjaxRequestNavData) && !requestNavData.equals(lastRequestNavData)) {
-                    NodeURL nodeURL = pcontext.createURL(NodeURL.TYPE).setNode(getCurrentSite().getSelectedUserNode());
-                    pcontext.sendRedirect(nodeURL.toString());
-                    return;
-                }
+          if (isAjaxInLastRequest) {
+            isAjaxInLastRequest = false;
+            if (requestNavData.equals(lastNonAjaxRequestNavData) && !requestNavData.equals(lastRequestNavData)
+                && pcontext.getPortletParameters().isEmpty()) {
```
ahamdi, line 823: Could you tell more about the relation between the request parameters and the redirection ? why did you add the check on the portlet parameters ?

ahamdi (approved):

PR #296, opened 2021-11-23
MAINT-48685: Avoid the display of a blank page when the csrf token of webui request event is expired

hakermi:
… 
**ISSUE**: When open the documents app and leave it for a about 30 minutes, the csrf token of the session will be updated and all the webui event requests from the opened page in document app is using the already old bound csrf tokens in the page which causing a false alert because the sent token in the request is not equal to the session token and so the request will be rejected and a blank page will be shown.
**SOLUTION**: Throw an exception to reload the page if this incident takes a place to avoid the blank page display

### DIFF
--- a/webui/framework/src/main/java/org/exoplatform/webui/event/Event.java
+++ b/webui/framework/src/main/java/org/exoplatform/webui/event/Event.java
@@ -25,6 +25,7 @@ import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.web.security.csrf.CSRFTokenUtil;
 import org.exoplatform.webui.application.WebuiRequestContext;
+import org.exoplatform.webui.exception.CSRFException;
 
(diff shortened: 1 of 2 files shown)


webui/framework/src/main/java/org/exoplatform/webui/event/Event.java
```diff
@@ -25,6 +25,8 @@
 import org.exoplatform.services.log.Log;
 import org.exoplatform.web.security.csrf.CSRFTokenUtil;
 import org.exoplatform.webui.application.WebuiRequestContext;
+import org.exoplatform.webui.exception.CSRFException;
+import org.exoplatform.webui.exception.MessageException;
```
ahamdi: unneeded import.

hakermi: removed

ahamdi (approved):

PR #301, opened 2021-12-08
MAINT-48685: Catch CsrfException and reload the page when the csrf token has been lost or expired

hakermi:
…
ISSUE: When the csrf token is lost and expired in the session, and we have alread opened pages of the doc app, This lead to a blank page when execute any action that demands a csrf token check
FIX: Reload the page when a csrf token is expired or lost to avoid the display of blank page and to refresh the token in the page

### DIFF
--- a/webui/framework/src/main/java/org/exoplatform/webui/core/UIApplication.java
+++ b/webui/framework/src/main/java/org/exoplatform/webui/core/UIApplication.java
@@ -27,6 +27,7 @@ import org.exoplatform.services.log.Log;
 import org.exoplatform.web.application.AbstractApplicationMessage;
 import org.exoplatform.web.application.ApplicationMessage;
 import org.exoplatform.webui.application.WebuiRequestContext;
+import org.exoplatform.webui.exception.CSRFException;
 import org.exoplatform.webui.exception.MessageException;
 
 /**
@@ -123,6 +124,8 @@ public abstract class UIApplication extends UIContainer {
             super.processAction(context);
         } catch (MessageException ex) {
             addMessage(ex.getDetailMessage());
+        } catch (CSRFException e) {
+            context.getJavascriptManager().getRequireJS().addScripts("location.reload();");
         } catch (Throwable t) {
             ApplicationMessage msg = new ApplicationMessage("UIApplication.msg.unknown-error", null, ApplicationMessage.ERROR);
             addMessage(msg);


webui/framework/src/main/java/org/exoplatform/webui/core/UIApplication.java
```diff
@@ -123,10 +124,12 @@ public void processAction(WebuiRequestContext context) throws Exception {
             super.processAction(context);
         } catch (MessageException ex) {
             addMessage(ex.getDetailMessage());
-        } catch (Throwable t) {
+        } catch (CSRFException e) {
+            context.getJavascriptManager().getRequireJS().addScripts("location.reload();");
+        } catch (Exception e) {
```
rdenarie:
```suggestion
        } catch (Trhowable t) {
```

webui/framework/src/main/java/org/exoplatform/webui/core/UIApplication.java
```diff
@@ -123,10 +124,12 @@ public void processAction(WebuiRequestContext context) throws Exception {
...
             ApplicationMessage msg = new ApplicationMessage("UIApplication.msg.unknown-error", null, ApplicationMessage.ERROR);
             addMessage(msg);
-            log.error("Error during the processAction phase", t);
+            log.error("Error during the processAction phase", e);
```
rdenarie:
```suggestion
            log.error("Error during the processAction phase", t);
```

rdenarie (changes requested):

webui/framework/src/main/java/org/exoplatform/webui/exception/CSRFException.java
```diff
@@ -1,6 +1,6 @@
 package org.exoplatform.webui.exception;
 
-public class CSRFException extends Exception{
+public class CSRFException extends RuntimeException{
```
rdenarie: Why did you change the exception type ? Is there a particular reason ?

rdenarie (changes requested):

webui/framework/src/main/java/org/exoplatform/webui/exception/CSRFException.java
hakermi: There is no major difference, but **RuntimeException** is an unchecked exception, and those exception are used when there is no recover for the thrown issue which is dealing with our situation where we cant' do anything except reloading the page when this exception comes to happen

hakermi: After thinking, I restore the Exception as we are handling it in a try catch clause

rdenarie (approved):

PR #305, opened 2021-12-21
MAINT-52380: Update error message when validating password policy

hakermi:
**ISSUE**: After updating the default validators for min=9 and max=256 length in password policy we need to update the error messages to indicate the right information for the user when he's about creating or updating a password.
**FIX**: This fix should update the error message by changing the new min length indicated in the message and hiding the max allowed length in the error message.


rdenarie (changes requested):
What about messages for 
- user settings (user change his password)
- user administration (user creation and edition by and administrator)
In addition, can you update description and commit message to follow the guideline ?

Thanks

hakermi:
> What about messages for
> 
> * user settings (user change his password)
> * user administration (user creation and edition by and administrator)
>   In addition, can you update description and commit message to follow the guideline ?
> 
> Thanks

All these cases are handled in the above fix, they all passes by the same rest endpoint and same validator 
For PR description and commit message, it seems okay for me (**why=> messages are not consistent because we updated validators**, **how=> we updated the error messages**), what should i change ?

rdenarie:
As explained in developer guide, and showed in the example template, the commit message must answer to the WHY question and the HOW question with 2 separated sentences for more consistency. Example : 

> Why is this change needed?
> Prior to this change, ...
> How does it address the issue?
> This change ...

Your actual commit message is very short, and in few weeks/month, if we had (or if someone else have to read your commit), it will be difficult to understand.
 
For example, it can be : 

> Before this fix, the default password length was 6 character min, and 20 characters max. Theses limitations makes the password weak regarding actual recommandations. 
> This fix change the default values in properties, to have at least 9 characters (which is defined as the "low to medium" strength by the ANSII password recommandations). 
> The recommandation also advise for not having maximum limit for the password. But to prevent potential attacks on that, we decide to put a huge limit to 256 characters. This upper limit will not be exposed to the user for security reason.


As you can see, the commit message, should not be just a small sentence, but a complete explaination of why you do this, and what was to mind path. So that, the person will read that in 1,2 or 6 month will be able to understand the problem in few seconds, and will not have to dig in task history to understand the problem.

hakermi:
> As explained in developer guide, and showed in the example template, the commit message must answer to the WHY question and the HOW question with 2 separated sentences for more consistency. Example :
> 
> > Why is this change needed?
> > Prior to this change, ...
> > How does it address the issue?
> > This change ...
> 
> Your actual commit message is very short, and in few weeks/month, if we had (or if someone else have to read your commit), it will be difficult to understand.
> 
> For example, it can be :
> 
> > Before this fix, the default password length was 6 character min, and 20 characters max. Theses limitations makes the password weak regarding actual recommandations.
> > This fix change the default values in properties, to have at least 9 characters (which is defined as the "low to medium" strength by the ANSII password recommandations).
> > The recommandation also advise for not having maximum limit for the password. But to prevent potential attacks on that, we decide to put a huge limit to 256 characters. This upper limit will not be exposed to the user for security reason.
> 
> As you can see, the commit message, should not be just a small sentence, but a complete explaination of why you do this, and what was to mind path. So that, the person will read that in 1,2 or 6 month will be able to understand the problem in few seconds, and will not have to dig in task history to understand the problem.

Okay i see the point now.

rdenarie (approved):

PR #306, opened 2021-12-21
Task-52379 : Fixed changed new password which is identical to the old one

Noussairhamrit:
ISSUE : Password can be change to the same one 
FIXED: when changing password, check if we are able to authenticate the user with the new password. If yes, it means that the user didnt change his password, so we refuse the modification


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -436,6 +438,10 @@ public Response changePassword(
       if (isSameUser && !userHandler.authenticate(username, currentPassword)) {
         return Response.serverError().entity(WRONG_USER_PASSWORD_ERROR_CODE).build();
       }
+
+      if (userHandler.authenticate(username, newPassword)==true) {
+          return Response.serverError().entity(WRONG_NEW_PASSWORD_ERROR_CODE).build();
+      }
```
```suggestion
      if (isSameUser && userHandler.authenticate(username, newPassword)) {
          return Response.serverError().entity(WRONG_NEW_PASSWORD_ERROR_CODE).build();
      }
```

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -43,6 +43,8 @@
 
   private static final String            DELEGATED_GROUP                = "/platform/delegated";
 
+  public static final String             WRONG_NEW_PASSWORD_ERROR_CODE = "WRONG_NEW_PASSWORD";
```
```suggestion
  public static final String             UNCHANGED_NEW_PASSWORD_ERROR_CODE = "UNCHANGED_NEW_PASSWORD";
```

rdenarie (changes requested):
About your commit message, can you develop the part "How you fix" ? 
In fact, in your solution, you do not compare the old and the new password, but you try to authenticate the user with the new password. 
It seems ok, but please change your description. 
In addition, can you follow the developer guide about commit message : 
https://developer.exoplatform.org/#id-contributions

Your commit message should contains 3 lines (at least) : 

> Task-52379 : prevent user to change password to the actual one
> 
> Before this fix ....
> This change ....

rdenarie (changes requested):
Please update your commit message as requested

rdenarie (approved):

PR #310, opened 2022-01-03
Task-49046 : adding user have the same name as deleted one

Noussairhamrit:
ISSUES : when adding user have the same name as deleted one  an error message is displayed, with an unknown error message
FIX: To resolve this issue I worked on two  projects gatein-portal and social so in gatein-portal firstly i add a new exception class 'ObjectAlreadyExistsException'  then i use  it in 'UserRestResourcesV1' with try ,catch   block and I did all that to throw the ObjectAlreadyExistsException exception at class SocialUserEventListenerImpl (social project ) instead of RuntimeException which is logically false and does not define the true execption


component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -222,7 +222,11 @@ public Response createUser(@Context HttpServletRequest request, UserRestEntity u
     user.setFirstName(firstName);
     user.setLastName(lastName);
     user.setPassword(password);
-    organizationService.getUserHandler().createUser(user, true);
+    try {
+       organizationService.getUserHandler().createUser(user, true);
+    } catch (ObjectAlreadyExistsException objectAlreadyExistsException) {
```
rdenarie, line 227: I think you dont have to create a new exception here. Catch the RuntimeException, and check the error message. If the error message is "Unable to create a previously deleted user", then you are in the case username already exists

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
```diff
@@ -222,7 +222,11 @@ public Response createUser(@Context HttpServletRequest request, UserRestEntity u
...
+       return Response.status(Response.Status.BAD_REQUEST).entity("USER:ALREADY_EXISTS_DELETED").build();
```
```suggestion
       return Response.status(Response.Status.BAD_REQUEST).entity("USERNAME:ALREADY_EXISTS_AS_DELETED").build();
```

rdenarie (changes requested):

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
Noussairhamrit: I see that it is not logical to throw RuntimeException in this case, again this exception returns status 500 and I cannot catch the error and with ObjectAlreadyExistsException I have a satuts 400 and i can catch the error message

abenmbarek: i think creating a new Exception class could help handling the right error specially from the front side using the right code error, he can throw 400 or 406 status error code in this case.
@rdenarie wdyt ?

rdenarie: Ok to keep the new Exception class

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
rdenarie: @Noussairhamrit can you make this modification ?

component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java
ahamdi: @rdenarie The idea is to use this Exception class and remove the one here https://github.com/Meeds-io/social/blob/develop/component/api/src/main/java/org/exoplatform/social/common/ObjectAlreadyExistsException.java
But I didn't want to we make this PR bigger and change in many places of the product. Maybe we could do it in another PR WDYT ?

rdenarie: can we keep this exception and modify the social one to extends this one ? 
And mark it as deprecated ?
So, all classes using social class will continue to work, and we can start to use this class instead ?

ahamdi: Seems good to me :+1:

rdenarie: @abenmbarek can you make the mdification in social module  ? Thanks

abenmbarek: no problem

rdenarie: @abenmbarek Im sorry, the message was for @Noussairhamrit which is the author of the PR :)

abenmbarek: ah it's okey, i'm here if he needs some help ;)

component/api/src/main/java/org/exoplatform/commons/ObjectAlreadyExistsException.java
```diff
@@ -0,0 +1,26 @@
+package org.exoplatform.commons;
```
rdenarie, line 17: Add copyright header please

rdenarie (changes requested):

rdenarie (approved):

PR #318, opened 2022-01-24
Task-52381: Token are not deleted after consumption.

montassarhamadi:
Problem: When user forgot his password, he when to reset it by sending a link to his mail which gives him the hand to change his password. So the problem has happened when he send many links to mail to reset the password, all links kept available to change his password Even if he did it the first time. all this happened because every time a link is sent to his mail a token is generated and stored automatically in the database, so the recent tokens keep stored even if he consumes one of them that's why we could reset the password by every link sent.

Fix: The solution is to remove all tokens by type (forgot-token, onboarding,external-registration) associate with this user after he succeeds to reset his password or create an account. I use the function deleteTokensByUsernameAndType(username, tokenType) of service RemindPasswordTokenService .this function is executed after every password is changed to delete all Tokens by type of the current user stored in DB then the recent links will be not available.


rdenarie (changes requested):
Backport not needed

To close

PR #319, opened 2022-01-25
MAINT-53372 : External-registration - user added to the users as default group and then removed and added to the externals group during the registration phase

hakermi:
…
ISSUE: When bind the users a group to a given space and create an external user the binding report will contain the external user as it's added to the users group and then removed, This because of broadcasting the `creation_event` to the `NewUserEventListener` which adds the user to the default `users` group.
Fix: This fix should disable the broadcast event option when creation the external user also remove the unneeded removing from the users group, So it will not be added only to the external group

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/externalRegistration/ExternalRegistrationHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/externalRegistration/ExternalRegistrationHandler.java
@@ -197,13 +197,9 @@ public class ExternalRegistrationHandler extends WebRequestHandler {
                       user.setEmail(email);
                     }
                     try {
-                        organizationService.getUserHandler().createUser(user, true);// Broadcast user creation event
+                        organizationService.getUserHandler().createUser(user, false);
                         Group group = organizationService.getGroupHandler().findGroupById(EXTERNALS_GROUP);
                         if (organizationService.getMembershipTypeHandler() != null) {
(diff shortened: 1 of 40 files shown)


component/web/security/src/main/java/org/exoplatform/web/login/externalRegistration/ExternalRegistrationHandler.java
```diff
@@ -197,13 +197,9 @@ public boolean execute(ControllerContext context) throws Exception {
                       user.setEmail(email);
                     }
                     try {
-                        organizationService.getUserHandler().createUser(user, true);// Broadcast user creation event
+                        organizationService.getUserHandler().createUser(user, false);
```
rdenarie, line 200: This action completly remove all users listeners. We should not remove

rdenarie (changes requested):

component/web/security/src/main/java/org/exoplatform/web/login/externalRegistration/ExternalRegistrationHandler.java
hakermi: yes but it's not needed for external users, and everything is working fine

hakermi: Ok, I have another solution for this.

rdenarie: I confirm that we need to broacast the event. When creating a user, you have 11 listeners registred :
image

chat, wiki, ES, ....
We cannot set broadcast to false

PR #321, opened 2022-01-28
MAINT-45289: Avoid having a null portalowner value in PortalRequestContext

hakermi:
**ISSUE**: Sometimes the notification of activities redirects to 404 error page, because of a null value of `portalOwner` in the URL.
**FIX**: THis fix should maintain the `getPortalOwner` function to avoid having such null value in the `portalOwner` by add a check and sending the default `portalOnwer` if the value is null

### DIFF
--- a/webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
+++ b/webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
@@ -511,10 +511,10 @@ public class PortalRequestContext extends WebuiRequestContext {
 
     public String getPortalOwner() {
         UserPortalConfig userPortalConfig = getUserPortalConfig();
-        if (userPortalConfig != null) {
+        if (userPortalConfig != null && userPortalConfig.getPortalName() != null) {
             return userPortalConfig.getPortalName();
         } else {
-            return null;
+            return portalConfigService.getDefaultPortal();
         }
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
This modification could have a huge impact as it is used in a lot of class of the platform.
Is it possible to isolate which part call this to generate the news notification, and make a specific treatement if getPortalOwner return null ?

I'm afraid of side effects with this current solution

hakermi:
> This modification could have a huge impact as it is used in a lot of class of the platform. Is it possible to isolate which part call this to generate the news notification, and make a specific treatement if getPortalOwner return null ?
> 
> I'm afraid of side effects with this current solution

I wanted to fix the source of the issue, it's juts a null check and return of default portalOwner which is DW
same treatment used here: https://github.com/Meeds-io/commons/blob/develop/commons-component-common/src/main/java/org/exoplatform/commons/utils/CommonsUtils.java#L191

hakermi:
> This modification could have a huge impact as it is used in a lot of class of the platform. Is it possible to isolate which part call this to generate the news notification, and make a specific treatement if getPortalOwner return null ?
> 
> I'm afraid of side effects with this current solution

@rdenarie  Could you check now ? i replaced it with the function in `commonUtils` which treats the null value

rdenarie (approved):

PR #322, opened 2022-02-02
TASK-53648: Allow to choose the Swedish language from language settings

abenmbarek:
Prior to this change, it was not possible to navigate in the portal with Swedish language, after this change we add a new local config related for Swedish language in order to allow navigating with Swedish language.


azayati (approved):
LGTM after QA validation

PR #323, opened 2022-02-04
TASK-53709: The new changed username isn't considered in chat application.

IlhemEssaadi:
ISSUES : When editing username, The new changed username isn't considered in chat application.
FIX : The problem is that when editing the firstname or the last name ,the new changes does not considered in chat application and it solved by adding the new displayname to the user object that will be updated in the mongo database.

### DIFF
--- a/component/portal/src/test/java/org/exoplatform/portal/rest/UserRestResourcesTest.java
+++ b/component/portal/src/test/java/org/exoplatform/portal/rest/UserRestResourcesTest.java
@@ -461,6 +461,20 @@ public class UserRestResourcesTest extends BaseRestServicesTestCase {
     assertEquals(400, response.getStatus());
     assertNotNull(response.getEntity());
     assertEquals("DisableSuperUser", response.getEntity().toString());
+
+    data.put("userName", USER_2);
+    data.put("lastName", USER_2);
+    data.put("firstName", USER_2);
+    data.put("password", "");
+    data.put("enabled", true);
+    data.put("email", email2);
+    response = getResponse("PUT", "/v1/users", data.toString());
+    assertNotNull(response);
+    assertNull(response.getEntity());
+    assertEquals(204, response.getStatus());
+    user2.setDisplayName(USER_2+" "+USER_2);
+    verify(userHandler, atMost(1)).saveUser(eq(user2), eq(true));
+
   }
 
   public void testDeleteUser() throws Exception {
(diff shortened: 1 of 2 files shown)


rdenarie (changes requested):
Can you add a little unit test to validate your modification ?
Thanks

component/portal/src/test/java/org/exoplatform/portal/rest/UserRestResourcesTest.java
```diff
@@ -461,6 +461,20 @@ public void testUpdateUser() throws Exception {
...
+    data.put("firstName", USER_2);
+    data.put("password", "");
+    data.put("enabled", true);
+    data.put("email", email2);
+    response = getResponse("PUT", "/v1/users", data.toString());
+    assertNotNull(response);
+    assertNull(response.getEntity());
+    assertEquals(204, response.getStatus());
+    user2.setDisplayName(USER_2+" "+USER_2);
+    verify(userHandler, atMost(1)).saveUser(eq(user2), eq(true));
```
rdenarie, line 476: Can you explain what this line verify ?

rdenarie (changes requested):
Can you fix the sonar feedback

IlhemEssaadi:
> Can you fix the sonar feedback

We can not fix them, they ask to remove the Mockito eq function which is required to pass the arguments. when I tried to remove one of them I got : 
```org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Invalid use of argument matchers!
2 matchers expected, 1 recorded:
-> at org.exoplatform.portal.rest.UserRestResourcesTest.testUpdateUser(UserRestResourcesTest.java:476)

This exception may occur if matchers are combined with raw values:
    //incorrect:
    someMethod(anyObject(), "raw String");
When using matchers, all arguments have to be provided by matchers.
For example:
    //correct:
    someMethod(anyObject(), eq("String by matcher"));
```

component/portal/src/test/java/org/exoplatform/portal/rest/UserRestResourcesTest.java
IlhemEssaadi: This line verify the save of the new user object after the update.

rdenarie:
And what if you put 
`verify(userHandler, atMost(1)).saveUser(user2, true);`

IlhemEssaadi:
> And what if you put `verify(userHandler, atMost(1)).saveUser(user2, true);`

The same problem!

rdenarie (approved):

PR #325, opened 2022-02-07
MAINT-45289: Avoid having a null portalowner value in PortalRequestContext

rdenarie:
ISSUE: Sometimes the notification of activities redirects to 404 error page, because of a null value of portalOwner in the URL.
FIX: THis fix should maintain the getPortalOwner function to avoid having such null value in the portalOwner by add a check and sending the default portalOnwer if the value is null


rdenarie:
cc @Montassar-Hamadi @wassimmhamdi @hakermi for validation

hakermi (approved):

PR #326, opened 2022-02-08
Task-53093: Allow to import new applications to an existing category

azayati:
Prior to this change, it was not possible to import new applications to an existing category. After this change, we have added a new "system" value param for ApplicationCategoriesPlugins component plugin in order to allow importing only new applications to an existing category and avoid reimporting already existing applications (when modified by interface).


hassenelkamel-exo (dismissed):
LGTM

boubaker (approved):

hassenelkamel-exo (approved):

PR #330, opened 2022-02-16
Task-54115 : No longer possible to create NEW customer spaces

Noussairhamrit:
ISSUES : error when creating customer space and an  error displayed  (An error occured when creating the space) 
FIX : the main cause of the bug is the deletion of a node. then in the creation of a space among the iteration there is a deletion of node children. at this level there an infinite loop is triggered because the deletion of node will remove data orphan of this node and in the same iteration this node there will be considered as an orphan and to solve this problem I removed the property 'orphanRemoval = true' at the level of the attribute children in 'NodeEntity


ahamdi (commented):
The problem is not yet fixed, this PR won't fix the issue as the changed function will return always null which will hide the right error as the delete opertaion of Description entity won't be executed.

rdenarie (approved):

PR #365, opened 2022-04-19
Task-56371 : Unable to login with alternative authentication system after few attempts

rdenarie:
Before this fix, there was a problem with alternative authentication system. After few successul attempts, the account is locked.
The problem is due to the fact the authencation fail counter is not reset on successful login.
To fix this problem, the OrganizationAuthenticatorImpl is moved to portal module, and the code to check the authentication attempts is moved from UserDAOImpl to OrganizationAuthenticatorImpl


component/api/src/main/java/org/exoplatform/services/organization/auth/AuthenticatorPlugin.java
```diff
@@ -0,0 +1,37 @@
...
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.exoplatform.services.organization.auth;
```
boubaker: I think It would be better to define this Service next to Organization Service, WDYT ?

pom.xml
```diff
@@ -39,8 +39,13 @@
   <name>GateIn - Portal</name>
   
   <properties>
+<<<<<<< HEAD
```
boubaker: Need to continue rebase

rdenarie: done

component/api/src/main/java/org/exoplatform/services/organization/auth/AuthenticatorPlugin.java
rdenarie: done

component/api/src/main/java/org/exoplatform/services/organization/auth/OrganizationAuthenticatorImpl.java
```diff
@@ -0,0 +1,453 @@
+/*
```
boubaker: Can you define this module inside component/identity to make it imported at the same time than Organization Service ?
The API Module doesn't hold any implementation and shouldn't  as you can see, it's made of constants, DTO, Interfaces, exceptions ...

rdenarie: done

component/api/pom.xml
```diff
@@ -71,6 +75,11 @@
       <artifactId>lombok</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
```
boubaker:
```suggestion
```
useless dependency

component/identity/src/test/java/org/exoplatform/services/organization/auth/TestOrganizationAuthenticator.java
```diff
@@ -0,0 +1,205 @@
+/*
+ * Copyright (C) 2009 eXo Platform SAS.
```
plamarque: AGPL 2022

boubaker (approved):

PR #399, opened 2022-07-06
Task 58776 : Update service parameter to use only one parameter wellKnowUrl

hchaabene:
**Prior to this change**, We need to validate the jwt token. For this we need information (keys) which are in the wellKnownUrl configuration , so we have rationalized and reduced the number of parameters.
**This change** , wellKnownUrl contains parameters userInfoUrl, TokenInfoUrl and authenticationURL  . By adding parameter wellKnowUrl, we are able to read other parameters, so we can remove them for cofnigruation.

### DIFF
--- a/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
+++ b/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
@@ -1,8 +1,11 @@
 package org.gatein.security.oauth.openid;
 
+import java.io.BufferedReader;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Base64;
 import java.util.HashMap;
@@ -10,8 +13,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-
-import javax.json.JsonObject;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
@@ -40,11 +41,10 @@ import org.picocontainer.Startable;
 
 /**
  * {@inheritDoc}
- *
  */
-public class OpenIdProcessorImpl implements OpenIdProcessor , Startable {
+public class OpenIdProcessorImpl implements OpenIdProcessor, Startable {
 
-  private static Log                log    = ExoLogger.getLogger(OpenIdProcessorImpl.class);
+    private static Log log = ExoLogger.getLogger(OpenIdProcessorImpl.class);
 
     private final String redirectURL;
     private String authenticationURL;
@@ -55,19 +55,15 @@ public class OpenIdProcessorImpl implements OpenIdProcessor , Startable {
     private final String clientSecret;
     private final Set<String> scopes = new HashSet<String>();
     private final String accessType;
-
     private final String wellKnownConfigurationUrl;
-    private String oidcServer;
     private final String applicationName;
     private final int chunkLength;
 
     private final SecureRandomService secureRandomService;
-    private final RemoteJwkSigningKeyResolver remoteJwkSigningKeyResolver;
 
     public OpenIdProcessorImpl(ExoContainerContext context, InitParams params, SecureRandomService secureRandomService) {
         this.clientID = params.getValueParam("clientId").getValue();
         this.clientSecret = params.getValueParam("clientSecret").getValue();
-        this.tokenInfoURL = params.getValueParam("tokenInfoURL").getValue();
         String redirectURLParam = params.getValueParam("redirectURL").getValue();
         this.wellKnownConfigurationUrl = params.getValueParam("wellKnownConfigurationUrl").getValue();
         String scope = params.getValueParam("scope").getValue();
@@ -78,11 +74,6 @@ public class OpenIdProcessorImpl implements OpenIdProcessor , Startable {
         } else {
             applicationName = "GateIn portal";
         }
-        ValueParam localOidcServer = params.getValueParam("oidcServer");
-        if (localOidcServer!=null) {
-            this.oidcServer = localOidcServer.getValue();
-        }
-
         if (clientID == null || clientID.length() == 0 || clientID.trim().equals("<<to be replaced>>")) {
             throw new IllegalArgumentException("Property 'clientId' needs to be provided. The value should be " +
                     "clientId of your OpenId application");
@@ -96,7 +87,7 @@ public class OpenIdProcessorImpl implements OpenIdProcessor , Startable {
 
         if (redirectURLParam == null || redirectURLParam.length() == 0) {
             this.redirectURL = "http://localhost:8080/" + context.getName() + OAuthConstants.OPEN_ID_AUTHENTICATION_URL_PATH;
-        }  else {
+        } else {
             this.redirectURL = redirectURLParam.replaceAll("@@portal.container.name@@", context.getName());
         }
 
@@ -115,12 +106,11 @@ public class OpenIdProcessorImpl implements OpenIdProcessor , Startable {
         }
 
         this.secureRandomService = secureRandomService;
-        this.remoteJwkSigningKeyResolver = new RemoteJwkSigningKeyResolver(this.oidcServer);
     }
 
     @Override
     public InteractionState<OpenIdAccessTokenContext> processOAuthInteraction(HttpServletRequest httpRequest,
-                                                                         HttpServletResponse httpResponse) throws IOException, OAuthException {
+                                                                              HttpServletResponse httpResponse) throws IOException, OAuthException {
         return processOAuthInteractionImpl(httpRequest, httpResponse, this.scopes);
     }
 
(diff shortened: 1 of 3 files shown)


component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +437,54 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
         return new URL(this.userInfoURL);
     }
+
+    @Override
+    public void start() {
+        if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
+            log.warn("wellKnownConfigurationUrl is not configured");
+            return;
+        }
+        JsonObject json = this.remoteJwkSigningKeyResolver.getJson(this.wellKnownConfigurationUrl);
```
rdenarie: Why do you use the remoteJwkSigningKeyResolver to read the configuration ?

The RemoteJwkSigningKeyResolver is used to get key to validate jwt token, not to read configuration

web/portal/src/main/webapp/WEB-INF/conf/sso/oauth-configuration.xml
```diff
@@ -155,21 +155,13 @@
...
-      </value-param>
-      <value-param>
-        <name>accessTokenURL</name>
-        <value>${exo.oauth.openid.accessTokenURL}</value>
-      </value-param>
       <value-param>
         <name>tokenInfoURL</name>
         <value>${exo.oauth.openid.tokenInfoURL}</value>
       </value-param>
       <value-param>
```
rdenarie: This should be removed also

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -73,6 +77,10 @@ public OpenIdProcessorImpl(ExoContainerContext context, InitParams params, Secur
         } else {
             applicationName = "GateIn portal";
         }
+        ValueParam localOidcServer = params.getValueParam("oidcServer");
+        if (localOidcServer != null) {
+            this.oidcServer = localOidcServer.getValue();
+        }
```
rdenarie: Not sure if we need it for this task. 
Perhaps when we make the third task about validating the jwt, but not needed in this current task

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -106,11 +114,12 @@ public OpenIdProcessorImpl(ExoContainerContext context, InitParams params, Secur
         }
 
         this.secureRandomService = secureRandomService;
+        this.remoteJwkSigningKeyResolver = new RemoteJwkSigningKeyResolver(this.oidcServer);
```
rdenarie: Not needed here. The signingKeyResolver will be use only for validate the jwt token. 
I dont think we need it in this task, perhaps in the tird one, but not here

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
```diff
@@ -64,7 +64,7 @@ public Key resolveSigningKey(JwsHeader header, String plaintext) {
     return getKey(header.getKeyId());
   }
 
-  private Key getKey(String keyId) {
```
rdenarie: I think we should not change here and keep private currently

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
```diff
@@ -86,7 +86,7 @@ private Key getKey(String keyId) {
     }
   }
 
-  private void updateKeys() {
```
rdenarie: I think we should not change here and keep private currently

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
```diff
@@ -113,7 +113,7 @@ private void updateKeys() {
 
   }
 
-  private JsonObject getJson(String url) {
```
rdenarie: I think we should not change here and keep private currently

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
```diff
@@ -135,7 +135,7 @@ private JsonObject getJson(String url) {
     }
   }
 
-  private BigInteger base64ToBigInteger(String value) {
```
rdenarie, line 138: I think we should not change here and keep private currently

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +425,75 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
     protected URL getUserInfoURL() throws IOException {
         if (log.isTraceEnabled())
             log.trace("UserInfo Request=" + this.userInfoURL);
         return new URL(this.userInfoURL);
     }
+
+    @Override
+    public void start() {
+        if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
+            log.warn("wellKnownConfigurationUrl is not configured");
```
rdenarie: I think we could put log.error here

rdenarie (changes requested):
Can you also try to fix the last sonar feedback ? It show you how to use try-with-resource, which automatically manage the reader.close()

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,80 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
+            String wellnownConfigurationContent = readUrl(this.wellKnownConfigurationUrl);
+            if (wellnownConfigurationContent != null) {
+                JSONObject json = new JSONObject(wellnownConfigurationContent);
+                this.authenticationURL = json.getString("authorization_endpoint");
+                this.accessTokenURL = json.getString("token_endpoint");
+                this.userInfoURL = json.getString("userinfo_endpoint");
+                this.tokenInfoURL = json.getString("token_endpoint");
+            }
+        } catch (JSONException e) {
+            log.error("error at read Url wellnown Configuration Content : " + e.getMessage());
```
```suggestion
            log.error("Unable to read webKnownUrl content.", e.getMessage());
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,80 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
+            return;
+        }
+        try {
+            String wellnownConfigurationContent = readUrl(this.wellKnownConfigurationUrl);
+            if (wellnownConfigurationContent != null) {
+                JSONObject json = new JSONObject(wellnownConfigurationContent);
+                this.authenticationURL = json.getString("authorization_endpoint");
+                this.accessTokenURL = json.getString("token_endpoint");
+                this.userInfoURL = json.getString("userinfo_endpoint");
+                this.tokenInfoURL = json.getString("token_endpoint");
```
rdenarie: This parameter shouldnt be use, we have to remove it
```suggestion
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,80 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
-}
+
+    @Override
+    public void start() {
+        if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
+            log.error("wellKnownConfigurationUrl is not configured");
+            return;
+        }
+        try {
+            String wellnownConfigurationContent = readUrl(this.wellKnownConfigurationUrl);
```
```suggestion
            String wellKnownConfigurationContent = readUrl(this.wellKnownConfigurationUrl);
```

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,80 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
+            log.error("Unable to read webKnownUrl content.", e.getMessage());
+        }
+    }
+
+    @Override
+    public void stop() {
+        // Nothing to stop
+    }
+
+    private static String readUrl(String urlString) {
```
```suggestion
    private static String readUrl(URL url) {
```
This will allow you to fix the last sonar feedback

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,76 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
+            if (wellKnownConfigurationContent != null) {
+                JSONObject json = new JSONObject(wellKnownConfigurationContent);
+                this.authenticationURL = json.getString("authorization_endpoint");
+                this.accessTokenURL = json.getString("token_endpoint");
+                this.userInfoURL = json.getString("userinfo_endpoint");
+            }
+        } catch (JSONException e) {
+            log.error("Unable to read webKnownUrl content.", e.getMessage());
+        } catch (MalformedURLException e) {
+            log.error(e.getMessage());
```
rdenarie: Can you add a real message explaining the problem ?

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -426,34 +428,76 @@ protected JSONObject parseResponse(String httpResponse) throws JSONException {
...
+            if (wellKnownConfigurationContent != null) {
+                JSONObject json = new JSONObject(wellKnownConfigurationContent);
+                this.authenticationURL = json.getString("authorization_endpoint");
+                this.accessTokenURL = json.getString("token_endpoint");
+                this.userInfoURL = json.getString("userinfo_endpoint");
+            }
+        } catch (JSONException e) {
+            log.error("Unable to read webKnownUrl content.", e.getMessage());
+        } catch (MalformedURLException e) {
+            log.error("Unknown wellKnownConfigurationUrl :" + e.getMessage());
```
rdenarie: Message should be something like 
"WellKnowConfigurationUrl malformed : url"
We should display the url in error log in order to be able to see why there is a problem

rdenarie (changes requested):
Can you also update the description ?
We should have a message explaining why we make this change, and how. The format should be something like : 
# Why is this change needed?
# Prior to this change, ...
# How does it address the issue?
# This change ...

web/portal/src/main/webapp/WEB-INF/conf/sso/oauth-configuration.xml
hchaabene: how i can get  tokenInfoURL  is not exist in .well-known/openid-configuration

rdenarie: TokenInfoUrl is used to validate the token.
As we change this to validate the token locally, we dont need it.
In addition, this url is not provided by all idp vendors

rdenarie (approved):

PR #400, opened 2022-07-11
Task-58777: Validate jwt token for OIDC

hchaabene:
Prior to this change, **tokeninfo** url for debugging only, and not in production mode. In production, the token validation should be done locally
This change , As we know have a key resolver, I update the function **validateTokenAndUpdateScopes**() with **Jwts**.parserBuilder()

### DIFF
--- a/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
+++ b/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
@@ -7,17 +7,14 @@ import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
 import com.github.scribejava.core.model.OAuth2AccessToken;
+import io.jsonwebtoken.*;
 import org.apache.commons.lang.StringUtils;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.services.log.ExoLogger;
@@ -49,7 +46,6 @@ public class OpenIdProcessorImpl implements OpenIdProcessor, Startable {
     private final String redirectURL;
     private String authenticationURL;
     private String accessTokenURL;
-    private String tokenInfoURL;
     private String userInfoURL;
     private final String clientID;
     private final String clientSecret;
@@ -57,8 +53,11 @@ public class OpenIdProcessorImpl implements OpenIdProcessor, Startable {
     private final String accessType;
     private final String wellKnownConfigurationUrl;
     private final String applicationName;
+    private String issuer;
     private final int chunkLength;
 
+
+    private RemoteJwkSigningKeyResolver remoteJwkSigningKeyResolver;
     private final SecureRandomService secureRandomService;
 
     public OpenIdProcessorImpl(ExoContainerContext context, InitParams params, SecureRandomService secureRandomService) {
@@ -237,14 +236,13 @@ public class OpenIdProcessorImpl implements OpenIdProcessor, Startable {
                     if (log.isTraceEnabled())
                         log.trace("Access Token=" + accessToken);
 
-                    OAuth2AccessToken tokenResponse = new OAuth2AccessToken(accessToken,
+                    return new OAuth2AccessToken(accessToken,
                             json.getString(OAuthConstants.TOKEN_TYPE_PARAMETER),
                             json.getInt(OAuthConstants.EXPIRES_IN_PARAMETER),
                             null,
                             json.has(OAuthConstants.SCOPE_PARAMETER) ?
                                     json.getString(OAuthConstants.SCOPE_PARAMETER) : null,
                             json.toString());
-                    return tokenResponse;
 
                 }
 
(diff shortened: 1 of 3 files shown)


component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +258,26 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
-            scopes = tokenInfo.has(OAuthConstants.SCOPE_PARAMETER) ? tokenInfo.getString(OAuthConstants.SCOPE_PARAMETER).split(" ") : null;
-        } catch (JSONException jsonException) {
-            throw new OAuthException(OAuthExceptionCode.ACCESS_TOKEN_ERROR,
-                    "Error during token validation: token format is ko");
+            jsonObject = new JSONObject(rawResponse);
+            String[] chunks = jsonObject.get("id_token").toString().split("\\.");
+            Base64.Decoder decoder = Base64.getUrlDecoder();
+            String payload = new String(decoder.decode(chunks[1]));
+            String aud = new JSONObject(payload).get("aud").toString();
+            String requireIssuer = new JSONObject(payload).get("iss").toString();
```
rdenarie: Why to you instanciate the JSONObject twice ? Can we do taht only once ?

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +258,26 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+            Jwts.parserBuilder()
+                    .setSigningKeyResolver(remoteJwkSigningKeyResolver)
+                    .requireIssuer(requireIssuer)
+                    .requireAudience(aud)
+                    .build().parseClaimsJws(jsonObject.get("id_token").toString())
+                    .getBody();
```
rdenarie: does parseClaims is necessary  ? does getBody is necessary ? can you extract id_token in a variable, so that you can use it instead of repeating the tostring

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +258,26 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+        } catch (MissingClaimException | IncorrectClaimException | JSONException e) {
+            log.error("the parsed JWT did not have the sub field");
```
rdenarie: Please separate exception case, and add a dedicated message to each

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +258,26 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
-                    "Error during token validation: token format is ko");
+            jsonObject = new JSONObject(rawResponse);
+            String[] chunks = jsonObject.get("id_token").toString().split("\\.");
+            Base64.Decoder decoder = Base64.getUrlDecoder();
+            String payload = new String(decoder.decode(chunks[1]));
+            String aud = new JSONObject(payload).get("aud").toString();
+            String requireIssuer = new JSONObject(payload).get("iss").toString();
+            Jwts.parserBuilder()
+                    .setSigningKeyResolver(remoteJwkSigningKeyResolver)
+                    .requireIssuer(requireIssuer)
```
rdenarie: you should not extract the issuer from the token. You must use the one from configuration

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +258,26 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+                    .requireAudience(aud)
```
rdenarie: same for audience. In fact, here, you extract something from the token, then check that it is equals to that : 

String audience = extractAudienceFromToken(token)
assertThat(audience == extractAudienceFromToken(token))

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,17 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
-            }
-            scopes = tokenInfo.has(OAuthConstants.SCOPE_PARAMETER) ? tokenInfo.getString(OAuthConstants.SCOPE_PARAMETER).split(" ") : null;
-        } catch (JSONException jsonException) {
-            throw new OAuthException(OAuthExceptionCode.ACCESS_TOKEN_ERROR,
-                    "Error during token validation: token format is ko");
+            Jwts.parserBuilder()
+                    .setSigningKeyResolver(remoteJwkSigningKeyResolver)
+                    .requireIssuer(this.issuer)
+                    .requireAudience(this.clientID)
+                    .build();
```
rdenarie: we need to parse the token. So, contrary to my first comment, instruction parseClaims is necessary :)
My bad

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,17 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+        } catch (MissingClaimException e) {
+            log.error("the parsed JWT did not have the sub field");
```
```suggestion
            log.error("Claims are missing in JWT Token",e);
```

rdenarie: What do you return in case of error ? we should not return the token as it worked.

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,17 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
         }
```
rdenarie, line 283: Exception IncorrectClaimException is missing, for example if issuer or aud do not correspond

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,23 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
-                    "Error during token validation: token format is ko");
+            String tokenId = new JSONObject(accessToken.accessToken.getRawResponse()).get("id_token").toString();
+            Jwts.parserBuilder()
+                    .setSigningKeyResolver(remoteJwkSigningKeyResolver)
+                    .requireIssuer(this.issuer)
+                    .requireAudience(this.clientID)
+                    .build().parseClaimsJws(tokenId)
+                    .getBody();
+        } catch (IncorrectClaimException e) {
+            log.error("issuer or aud does not equal the required", e.getMessage());
```
```suggestion
            log.error("Issuer or audience have not the correct value in the token", e);
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,23 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+        } catch (MissingClaimException e) {
+            log.error("Claims are missing in JWT Token", e);
```
```suggestion
            log.error("Required claims (issuer or audience) are missing in JWT Token", e);
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -261,88 +259,23 @@ protected OAuth2AccessToken parseResponse(String httpResponse) throws JSONExcept
...
+        } catch (JSONException e) {
+            log.error("can't get token_id in JSONObject");
```
```suggestion
            log.error("Unable to parse the accessToken",e);
```

rdenarie (changes requested):
In the 3 catches, you should have a return.
return null; or something like this to not return a OpenIdAccessTokenContext when the hwt validation fails. In addition, you should manage the case of the object return is null in function which call validateTokenAndUpdateScopes

rdenarie (approved):

PR #422, opened 2022-09-12
fix: OIDC properties : Wellknown url isn't well extracted - EXO-59697 - meeds-io/meeds#288

rdenarie:
Before this fix, during the OIDC protocol, we try to validate the JWT token For that, we use RemoteJwkSigningKeyResolver, and we initialize it with the issuer parameter. This class then rebuild the well-known url by concatenate issuer and "/.well-known/openid-configuration". When issuer is different from the well-know path, the class RemoteJwkSigningKeyResolver is unabel to find the well-known configuration file

This fix modify RemoteJwkSigningKeyResolver to pass in parameter the well-known url instead of the issuer

### DIFF
--- a/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
+++ b/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
@@ -38,13 +38,13 @@ import java.util.*;
 
 public class RemoteJwkSigningKeyResolver implements SigningKeyResolver {
 
-    private final String issuer;
+  private final String     wellKnownUrl;
     private final Object lock = new Object();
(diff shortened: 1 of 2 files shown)


component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/RemoteJwkSigningKeyResolver.java
```diff
@@ -38,13 +38,13 @@
 
 public class RemoteJwkSigningKeyResolver implements SigningKeyResolver {
 
-    private final String issuer;
+  private final String     wellKnownUrl;
```
azayati:
```suggestion
  private final String     wellKnownUrl;
```
indentation

azayati (changes requested):
see review

Jihed525 (approved):

azayati (approved):
LGTM

PR #425, opened 2022-09-16
fix: Fix blank page displayed when login with non existing email - EXO-59754 - Meeds-io/meeds#298

azayati:
Prior to this change, a blank page is displayed when trying to login with a non existing email. After this commit, we ensure to send failed authentication status when the findUsersByQuery with the taped email not returns users.

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
@@ -254,6 +254,10 @@ public class LoginHandler extends JspBasedWebHandler {
         if (users != null && users.getSize() > 0) {
           return users.load(0, 1)[0].getUserName();
         }
+        else {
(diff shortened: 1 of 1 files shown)


akhanfir (approved):

component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
```diff
@@ -254,6 +254,10 @@ private String getUserNameByEmail(String email,
         if (users != null && users.getSize() > 0) {
           return users.load(0, 1)[0].getUserName();
         }
+        else {
+          LOG.warn("Can not get users by email");
```
boubaker, line 258: It's better to avoid adding such information as Warning in log since it's a normal behavior and not an unexpected behavior from system. (you can add a `log.debug` if useful for troubleshooting)

PR #428, opened 2022-09-23
fix: Account is locked when password is changed - EXO-59903 - meeds-io/meeds#311

rdenarie:
Before this fix, when a user have a tab opened with an active remember token, and changing his password, the rememberme token is used to try authentication But, as the password was changed, the authentication fails at multiple times, which lead to lock the account

This fix deletes from tokenStore all rememberme token of the users, as they are all no more valid, so that when the tab is refresh with an old rememberme token, it is not found by the server, and so ignored. Then the login form is displayed without locking the account.


azayati (dismissed):
LGTM

azayati (changes requested):
please fix sonar checks

azayati (approved):
LGTM

PR #430, opened 2022-09-30
feat: Fix list of group collaborators' suggestions in the access management - EXO-59958 - Meeds-io/MIPs#23

IlhemEssaadi:
Prior to this change, the groups proposed in the list of collaborators' suggestions are only the groups of which I am a member. After this change, if I am an administrator with any role, all groups are suggested and if I am a user, only the groups I am a member of.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/ExtGroup.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/ExtGroup.java
@@ -29,13 +29,25 @@ import org.exoplatform.services.organization.impl.GroupImpl;
  * @author <a href="mailto:boleslaw.dawidowicz at redhat.com">Boleslaw Dawidowicz</a>
  */
 public class ExtGroup extends GroupImpl implements Comparable<Object> {
+
   private static final long serialVersionUID = -7379104016885124958L;
+  
+  private String            groupType;
 
   public ExtGroup() {
   }
-
+  
   public ExtGroup(String name) {
     super(name);
+  }
+
+  public ExtGroup(String name, String groupType) {
+    super(name);
+    this.groupType = groupType;
+  }
+
+  public String getGroupType() {
+    return groupType;
   }
 
   public String toString() {
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/ExtGroup.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/ExtGroup.java
@@ -32,8 +32,6 @@ public class ExtGroup extends GroupImpl implements Comparable<Object> {
 
   private static final long serialVersionUID = -7379104016885124958L;
   
-  private String            groupType;
-
   public ExtGroup() {
   }
   
@@ -41,14 +39,6 @@ public class ExtGroup extends GroupImpl implements Comparable<Object> {
     super(name);
   }
 
-  public ExtGroup(String name, String groupType) {
-    super(name);
-    this.groupType = groupType;
-  }
-
-  public String getGroupType() {
-    return groupType;
-  }
 
   public String toString() {
     return "Group[" + getId() + "|" + getGroupName() + "]";
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
@@ -21,10 +21,7 @@ package org.exoplatform.services.organization.idm;
 
 import java.io.IOException;
 import java.util.*;
-import java.util.stream.Collectors;
-
 import org.apache.commons.lang3.StringUtils;
-import org.exoplatform.services.security.Identity;
 import org.picketlink.idm.api.Attribute;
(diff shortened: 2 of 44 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -563,17 +571,31 @@ public Collection<Group> findGroupsOfUserByKeyword(String user, String keyword,
         return null;
       }
       Collection<org.picketlink.idm.api.Group> allGroups = new HashSet<org.picketlink.idm.api.Group>();
-      try {
-        orgService.flush();
-        allGroups = getIdentitySession().getRelationshipManager().findRelatedGroups(user, groupType, identitySearchCriteria);
-      } catch (Exception e) {
-        // TODO:
-        handleException("Identity operation error: ", e);
+      Identity identity = ConversationState.getCurrent().getIdentity();
```
mkrout: You can't change the behavior of this method here, it can be already used by other services that need the list of groups of a user no matter if he is an admin or not, your needs should be handled in the suggestion service, not in the DAO layer

mkrout (changes requested):

component/identity/src/main/java/org/exoplatform/services/organization/idm/ExtGroup.java
```diff
@@ -30,14 +30,20 @@
...
 public class ExtGroup extends GroupImpl implements Comparable<Object> {
   private static final long serialVersionUID = -7379104016885124958L;
+  
+  private String            groupType;
 
   public ExtGroup() {
   }
 
-  public ExtGroup(String name) {
+  public ExtGroup(String name, String groupType) {
```
mkrout: will be better if you add a new contructor, with new group type param

mkrout (approved):

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
```diff
@@ -573,7 +577,7 @@ public Collection<Group> findGroupsOfUserByKeyword(String user, String keyword,
       List<Group> exoGroups = new LinkedList<Group>();
       for (org.picketlink.idm.api.Group group : allGroups) {
         try {
-          if (groupType.isEmpty() || group.getGroupType().equals(groupType)) {
+          if (groupType.isEmpty() || group.getGroupType().equals(groupType) || !group.getGroupType().equals(SPACE_GROUP_TYPE) && !group.getGroupType().equals(ROOT_GROUP_TYPE)) {
```
rdenarie: I dont understand why we filter to not get groups with type SPACE_GROUP_TYPE and groups with type ROOT_GROUP_TYPE
As this function should return groups of users, if user is present in theses groups we must return it. We cannot remove arbitrary groups when someone request groups of users.

component/portal/src/main/java/org/exoplatform/portal/rest/GroupRestResourcesV1.java
```diff
@@ -641,25 +646,37 @@ public Response getGroupsTreeMembers(@Context
...
     limit = limit > 0 ? limit : DEFAULT_LIMIT;
-
-    Collection<Group> UserGroupsList = organizationService.getGroupHandler()
-                                                          .findGroupsOfUserByKeyword(identity.getUserId(), q, groupType);
-    int totalSize = UserGroupsList.size();
-    int limitToFetch = limit;
-    if (totalSize < (offset + limitToFetch)) {
-      limitToFetch = totalSize - offset;
-    }
+    CollectionEntity<Group> result = null;
```
rdenarie: What is the initial aim of this endpoint treeMembers ?

I'm afraid that we are changing the purpose of this endpoint to add a special case for administrators.

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
rdenarie: Instead of hardcoding some grouptype to exclude, should we add in the rest service a parameter named excludedGroupTypes ?

rdenarie (changes requested):
I think we need a little tech spec about theses change as they impact API

component/identity/src/main/java/org/exoplatform/services/organization/idm/GroupDAOImpl.java
mkrout: @IlhemEssaadi you should not filter the groups using the hardcoded type for spaces and should only return the groups with the type given in the method param: groupType.
 @rdenarie we need to filter with ROOT_GROUP_TYPE because we don't need to get root groups in the suggestor as for example (/PLatform, ...)

component/portal/src/main/java/org/exoplatform/portal/rest/GroupRestResourcesV1.java
mkrout: This endpoint was added for process development to add groups to the suggester, I don't think that it can be used otherwhere. but we will add a new query param to this endpoint to not change it behavior if no need to get all groups for admins.

PR #447, opened 2022-11-07
feat: Left menu review & adds MEED-801- Meeds-io/MIPs#16

### LINKED ISSUE #16: Left menu review
## Rationale

When using the platform, you have a left menu accessible from the burger menu.
This left menu offers the user access to generic pages like, stream, people, spaces, recent spaces, settings and administration when having administrator role
By default, the menu is collapsed and the user needs to open it to display the list. Then, it collapsed back.
From this menu, we can open second level menu like spaces of the current user and administration when being administrator.

Since it has been proposed to users, we collected feedbacks and we brought new options including for spaces that we need to adress to make the platform browsing smooth and smart.

## 1. Functional requirements
- Firstly, we'd like to review the way we can access to spaces to be clearer and consistent with the platform: 
** from the left nav, the arrow to display second panels (recent spaces or admin) are not displayed by default. The user needs to hover the line to see it and then he can click to it to display the second panel. This to avoid opening the panel while the user doesn't want to see it. (using mobile, you can click to the whole line)
** position the "add a space" access from the left nav directly to be consistent with other tools using this way to create spaces / communities. That way, the recent spaces panel proposes only a way to filter by space
** when the second panel is opened, the arrow (and the plus item for recent spaces) is remained displayed, the user can click to it to close the second panel
access to second panel

access to second panel mobile

- Then, for any space listed in the main left nav or in the recent spaces panel, the user will have access to more options so he can directly access an app or so he can manage to do an action: bookmark it for example (mobile or desktop)

Items listed in the space details panel depends of:
* actual items listed in the space navigation (sort, name)
* the permissions to the page (i.e settings only for space hosts)
* any new item added from the settings in a space should be listed depending the name, the sort and permissions

space details panel

space detail panels from recent spaces

space detail panel - mobile view

#### 2.1. Performance requirements
Knowing that navigation menu is displayed in all pages, it should be as much performant as before when changing the current new UI behaviors. Thus, the pre-loaded urls has to be updated when changing the fetched resources by UI improvements made.

#### 2.2. Feature Flag
* Feature flag name: `exo.feature.leftMenuReview.enabled`
* Default value: true
* By disabling this feature flag, the Left menu will be reset to use the Legacy Old Behavior without any improvement made by the current MIP.
* To set a target audience, you can use `exo.feature.leftMenuReview.permissions` to select a group of users.

#### 2.3. Security requirements (Authentication, Authorizations, Audit track / traceability, data retention…)
No addition security context is required to the existing one, which is:
- Display only authorized pages in Menu
- Display only authorized and accessed spaces in Menu
- Display administration menu only if user can access to one or multiple administrative pages

## 3. Impacts
- Documentation to update

#### Access (GUI, API…)
No new REST API endpoints will be added.

#### Services & processing
No change on Service & Storage layer will be made by this improvement.

SaraBoutej:
Prior to this change, when changing through XML the navigation nodes **order**, it's not impacted even with importMode = 'MERGE'.
The problem is caused by the fact that the `index` (position order in navigation tree) of the navigation node isn't updated by following the same `index` defined in navigation XML file.
This change allows to change the navigation nodes orders by XML configuration file by applying the exact `index` of nodes as defined in `navigation.xml` file.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/importer/NavigationFragmentImporter.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/importer/NavigationFragmentImporter.java
@@ -177,17 +177,24 @@ public class NavigationFragmentImporter {
         }
 
         // Buffer the changes in a list
-        LinkedList<Change> foo = new LinkedList<Change>();
+        LinkedList<Change> changes = new LinkedList<>();
         while (it.hasNext()) {
             ListChangeType type = it.next();
-            foo.add(new Change(type, it.getElement(), it.getIndex1(), it.getIndex2()));
+            changes.add(new Change(type, it.getElement(), it.getIndex1(), it.getIndex2()));
         }
+        changes.sort((change1, change2) -> {
+          PageNode srcNode1 = src.getNode(change1.name);
+          PageNode srcNode2 = src.getNode(change2.name);
+          int index1 = srcNode1 == null ? Integer.MAX_VALUE : src.getNodes().indexOf(srcNode1);
+          int index2 = srcNode2 == null ? Integer.MAX_VALUE : src.getNodes().indexOf(srcNode2);
+          return index1 - index2;
+        });
 
         // The last encountered child
         NodeContext<?> previousChild = null;
 
         // Replay the changes and apply them
-        for (Change change : foo) {
+        for (Change change : changes) {
             PageNode srcChild = src.getNode(change.name);
             NodeContext<?> dstChild = dst.get(change.name);
 
@@ -199,7 +206,7 @@ public class NavigationFragmentImporter {
 
                     //
                     if (config.updatedSame) {
-                        update(srcChild, dstChild, labelMap);
+                        update(src, dst, srcChild, dstChild, labelMap);
                     }
 
                     //
@@ -216,7 +223,7 @@ public class NavigationFragmentImporter {
                 case ADD:
                     if (src.getNode(change.name) != null) {
                         if (config.updatedSame) {
-                            update(srcChild, dstChild, labelMap);
+                            update(src, dst, srcChild, dstChild, labelMap);
                         }
                         previousChild = dstChild;
                     } else {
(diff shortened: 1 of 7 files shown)


rdenarie (changes requested):
Can you explain briefly what is the root cause ?
You answer question WHAT ? but not the question WHY ?
In addition can you answer the question HOW ? you fix it ?

It will help to understand the problem. 
I'm not sure currently to understand your fix.
Thanks

boubaker:
> Can you explain briefly what is the root cause ? You answer question WHAT ? but not the question WHY ? In addition can you answer the question HOW ? you fix it ?
> 
> It will help to understand the problem. I'm not sure currently to understand your fix. Thanks

Done

boubaker:
In this changes, two code snippets to read for the fix:

```java
        changes.sort((change1, change2) -> {
          PageNode srcNode1 = src.getNode(change1.name);
          PageNode srcNode2 = src.getNode(change2.name);
          int index1 = srcNode1 == null ? Integer.MAX_VALUE : src.getNodes().indexOf(srcNode1);
          int index2 = srcNode2 == null ? Integer.MAX_VALUE : src.getNodes().indexOf(srcNode2);
          return index1 - index2;
        });
```

Which will **sort** the **changes** to apply on existing navigation nodes to start from the first node of the tree as defined in `navigation.xml` to the last one.

Second code snippet:
```java
        int srcIndex = srcParent.getNodes().indexOf(src);
        if (srcIndex <= 0) {
          dstParent.add(0, target);
        } else {
          PageNode previousSrcNode = srcParent.getNodes().get(srcIndex - 1);
          NodeContext previousTargetNode = dstParent.get(previousSrcNode.getName());
          if (previousTargetNode != null) {
            dstParent.add(previousTargetNode.getIndex() + 1, target);
          } else {
            dstParent.add(null, target);
          }
        }
```

Which will place the source navigation node in storage exactly in :
* First position of tree if it's defined in XML as first element => `(srcIndex <= 0)`
* Just after the previous node defined in XML file. Example: In storage we had (A, C, B) and in XML (A, B, C), when it comes to 'B' navigation change to apply, it will be moved just after 'A', as it's defined in XML (Storage result: A, B, C). For 'C', it will be idempotent because the 'C' is already after 'B' in storage.

rdenarie (approved):

PR #455, opened 2022-11-10

IlhemEssaadi:
Prior to this change, it is possible to create an external user with first name and last name containing illegal eXo characters which generates the username also with illegal eXo characters and may make some problems. After this change, we should forbid these illegal eXo characters for the fields first name and last name when crating an external user by applying the same validator of username field when creating a simple user.


azayati (approved):
Just we need to format this file correctly in next develop PR

PR #461, opened 2022-11-15

IlhemEssaadi:
Prior to this change, in the PR https://github.com/Meeds-io/gatein-portal/pull/457 we have used the first and last name validators in order to validate external user creation and forbid illegal chars but it was always possible to create an external user with a first name or a last name containing the character "@" at the end. After this change, to fix that we will use the username validator instead of first and last name validators for external user creation validation.


Jihed525 (approved):

azayati (approved):
LGTM

PR #466, opened 2022-11-18
feat: Broadcast file service action events - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change, No events or listener are handling file service actions. 
This PR should broadcast fileservice generic events to allow managing them inside listeners

### DIFF
--- a/component/file-storage/src/main/java/org/exoplatform/commons/file/services/impl/FileServiceImpl.java
+++ b/component/file-storage/src/main/java/org/exoplatform/commons/file/services/impl/FileServiceImpl.java
@@ -11,6 +11,7 @@ import org.exoplatform.commons.file.services.FileService;
 import org.exoplatform.commons.file.services.FileStorageException;
 import org.exoplatform.commons.file.services.NameSpaceService;
 import org.exoplatform.commons.file.storage.DataStorage;
+import org.exoplatform.services.listener.ListenerService;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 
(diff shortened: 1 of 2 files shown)


component/file-storage/src/main/java/org/exoplatform/commons/file/services/util/Utils.java
```diff
@@ -0,0 +1,17 @@
+package org.exoplatform.commons.file.services.util;
```
rdenarie: Is it mandatory to have utils class to call listenerService ?
I think you should call it directly, without creating utils class

rdenarie (changes requested):

component/file-storage/src/main/java/org/exoplatform/commons/file/services/util/Utils.java
hakermi: Well, i just wanted to not duplicate try/catch bloc

rdenarie: I prefer to duplicate the try catch block, that allow us to make different error log messages, instead of adding a new Utils class. I find it more readable.

hakermi: Ok

component/file-storage/src/main/java/org/exoplatform/commons/file/services/impl/FileServiceImpl.java
```diff
@@ -120,7 +144,11 @@ public FileInfo deleteFile(long id) {
     FileInfo fileInfo = dataStorage.getFileInfo(id);
     if (fileInfo != null) {
       fileInfo.setDeleted(true);
-    }
+      try {
+        listenerService.broadcast(FILE_DELETED_EVENT, fileInfo, null);
+      } catch (Exception e) {
+        LOG.error("Error while broadcasting event: {}", FILE_DELETED_EVENT, e);
+      }    }
```
rdenarie: Should this part be after the real delete (dataStorage.updateFileInfo) ?

rdenarie (changes requested):

component/file-storage/src/main/java/org/exoplatform/commons/file/services/impl/FileServiceImpl.java
hakermi: done

rdenarie (approved):

PR #473, opened 2022-11-28
feat: Broadcast file service action events - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change, No events or listener are handling file service actions.
This PR should broadcast fileservice generic events to allow managing them inside listeners


boubaker (dismissed):

boubaker (approved):
LGTM 2

PR #497, opened 2022-12-26
feat: Move oAuth to Social to avoid overrides and complicated maintenance & evolutions - MEED-1308

boubaker:
This change will move all oAuth artifacts in one single artifact in social to avoid overrides and useless implementations and compiled sources. In fact, more we have classes inside an API, more the security breach is high and more the maintenance and evolution is complicated. In addition, a Mocked InMemory IDM implementation has been added to improve building performances when testing upper layer modules.


AzmiTouil (approved):

exo-swf:

exo-swf:

PR #518, opened 2023-01-11
feat: Remove the administration application registry category and the IFramePortlet application from tools application registry category - EXO-61480- Meeds-io/MPIs#37

sofyenne:
This change is going to remove the administration application registry category and the IFramePortlet application from tools application registry category.


azayati (approved):
LGTM

PR #520, opened 2023-01-16
fix: include AuthenticationPlugin inside the authentication transaction - Meeds-io/meeds#553

### LINKED ISSUE #553: AuthenticationPlugins don't close IDM transactions
**Steps to reproduce**

When many users authenticate in the same time, many transactions are opened and keeped open and may exceed the number of allowed DB connections in the pool.

**Current behaviour**

This creates blocked threads and server does no more respond user requests 
**Expected behavior**

The transactions should be closed correctly to avoid saturation on DB connection usage.

ahamdi:
Priori to this fix, authenticationPlugins are not included in the request lifecycle of the authentication and will open DB transaction that won't be closed.
The fix reintegrates the Success/Fail checks of the Authentication plugin inside the RequestLifecycle.

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/auth/OrganizationAuthenticatorImpl.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/auth/OrganizationAuthenticatorImpl.java
@@ -185,6 +185,33 @@ public class OrganizationAuthenticatorImpl implements Authenticator
            }
            // No exception occurred
            lastExceptionOnValidateUser.remove();
+           if (!success) {
+              for (AuthenticatorPlugin plugin : getPlugins()) {
+                 try {
+                    String validatedUserName = plugin.validateUser(credentials);
+                    if (StringUtils.isNotBlank(validatedUserName)) {
+                       success = true;
+                       username = validatedUserName;
+                       break;
+                    }
+                 } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
This fix is not ok. 
If the securityCheckAuthenticationPlugin throws an exception in doCheck (like https://github.com/exoplatform/anti-bruteforce/blob/develop/anti-bruteforce-services/src/main/java/org/exoplatform/antibruteforce/plugin/AntiBruteForceAuthenticationPlugin.java#L109), we will go in the associate catch, and the plugin.onCheckFail will be not executed.
This instruction, for anti brute force, count the fail attemps

rdenarie:
> This fix is not ok. If the securityCheckAuthenticationPlugin throws an exception in doCheck (like https://github.com/exoplatform/anti-bruteforce/blob/develop/anti-bruteforce-services/src/main/java/org/exoplatform/antibruteforce/plugin/AntiBruteForceAuthenticationPlugin.java#L109), we will go in the associate catch, and the plugin.onCheckFail will be not executed. This instruction, for anti brute force, count the fail attemps

This assumption was an error. 
In the catch, we have a throw which make the end the function not executed, so the initial version of the PR is better

rdenarie (approved):

PR #527, opened 2023-01-25
fix: Fix Triggering RememberMe filter on static resources - MEED-1597- Meeds-io/meeds#592

### LINKED ISSUE #592: High IDM Pool usage when login users with RememberMe Tokens
**Steps to reproduce**

1. Login with user by enabling "Remember Me" checkbox
3. Change system date to make the `rememberMe` token invalid (+7d)
4. Restart server
6. Display login page by disabling cache

**Current behaviour 1**

Error in server log while starting:

```
javax.persistence.RollbackException: Error while committing the transaction
	at org.hibernate.internal.ExceptionConverterImpl.convertCommitException(ExceptionConverterImpl.java:81)
	at org.hibernate.engine.transaction.internal.TransactionImpl.commit(TransactionImpl.java:104)
	at org.exoplatform.commons.persistence.impl.ExoTransactionalAspect.around(ExoTransactionalAspect.java:100)
	at org.exoplatform.web.security.jpa.JPAGateInTokenStorage.cleanExpired(JPAGateInTokenStorage.java:79)
	at org.exoplatform.web.security.security.CookieTokenService.cleanExpiredTokens(CookieTokenService.java:255)
```

**Current behaviour 2**

Multiple errors when displaying login page of type `authentication failed`

**Expected behavior**

1. The outdated cookie are cleared without exceptions while starting server
2. Avoid login user when serving static resources

boubaker:
Prior to this change, the scripts, css and rest calls triggers the RememberMeFilter when displaying login page. The RememberMefilter has to be triggered only on sites pages display to login the user when he has a cookie with name 'rememberme'. This change will at the same time fix the startup of CookieTokenService Based Services which will attempts to delete all at the same time all outdated tokens, which leads to some services to triggers an exception because the change has been already made asynchronously by other services inheriting from CookieTokenService. The fix is to execute by each service a 'DELETE' query on database instead of retrieving the entities and delete them one by one.

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/RememberMeFilter.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/RememberMeFilter.java
@@ -20,10 +20,14 @@
 package org.exoplatform.web.login;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 import javax.servlet.*;
 import javax.servlet.http.*;
 
+import org.apache.commons.lang3.StringUtils;
 import org.gatein.wci.ServletContainer;
 import org.gatein.wci.ServletContainerFactory;
 import org.gatein.wci.security.Credentials;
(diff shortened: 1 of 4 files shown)


component/web/security/src/main/java/org/exoplatform/web/login/RememberMeFilter.java
```diff
@@ -48,14 +52,28 @@ public class RememberMeFilter extends AbstractFilter {
     //value of this field need equals with: org.gatein.security.oauth.common.OAuthConstants.ATTRIBUTE_AUTHENTICATED_PORTAL_USER_FOR_JAAS
     public static final String ATTRIBUTE_AUTHENTICATED_PORTAL_USER_FOR_JAAS = "_authenticatedPortalUserForJaas";
 
+    public List<String>      ignoredPaths                                 = null;
+
+    @Override
+    protected void afterInit(FilterConfig config) throws ServletException {
+      String ignoredPathsParameter = config.getInitParameter("ignoredPaths");
+      if (StringUtils.isBlank(ignoredPathsParameter)) {
```
rdenarie, line 60: Shoud we use 
```suggestion
      if (ignoredPathsParameter==null || ignoredPathsParameter.isBlank()) {
```

to prevent using StringUtils library ?

rdenarie (dismissed):

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeFilter.java
boubaker: why preventing using commons library ?

rdenarie: To use standard java way. as java provide the possibility to check strings, is it correct to use external lib for that ? 

It is only a reflexion, not an objection, I should have put it in comment instead of Change Request

ahamdi (approved):

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeFilter.java
boubaker: Ok, let's discuss about this basic and general concern. In fact, what is made here is the purpose of third party libraries making the code more lisible and centralize the same concern in a re-usable way, even Spring has defined their own StringUtils class.

PR #532, opened 2023-02-01
fix: Trim the username/email field to login - EXO-61937

GouadriaHanen:
before this change, when a username or email address contains a space  in the beginning/end  (ex: "  root"), the login action is not executed and an error is displayed
after this change, spaces are cleaned (trim() for the username) to send the correct value for the authentication.

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
@@ -156,6 +156,9 @@ public class LoginHandler extends JspBasedWebHandler {
     }
 
     String username = request.getParameter("username");
+    if (username != null) {
+      username = username.trim();
+    }
     String password = request.getParameter("password");
 
     final String portalContextPath = servletContext.getContextPath();


component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
```diff
@@ -156,6 +156,9 @@ public boolean execute(ControllerContext context) throws Exception { // NOSONAR
     }
 
     String username = request.getParameter("username");
+    if (username != null) {
```
ahamdi, line 159: this check is not needed

ahamdi (changes requested):
Could you be more precise about the description of the problem ? and how it is fixed

component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
GouadriaHanen: needed in the case of loading the login page! the username and the password are null

ahamdi: Why isn't there a check on the password field ?

GouadriaHanen: no need to clean spaces in the password

ahamdi (approved):

PR #536, opened 2023-02-15
fix: Error in log when migrate from 1.2.5 to 1.3.x with psql database - EXO-62553 - meeds-io/meeds#692

rdenarie:
Before this fix, there was an error when performing migration from meed 1.2.5 to meeds 1.3.x with psql database This pb is due to liquibase update version in 1.3.x. With this update, blob type are now created in psql with oid type instead of bytea type When starting a fresh 1.3.x the type is correctly oid, so no problem to read it When performing a migration from 1.2.5, the type is still bytea and is not read correctly. The impact is on table PORTAL_WINDOWS, column customization

This fix add a liquibase update, which will be executed only if the type of this column is bytea. The update create a new temporary table, copy content into, and convert the bytea field to oid. Then drop the original table, then rename the temporary table with the original name.

### DIFF
--- a/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
+++ b/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
@@ -329,4 +329,18 @@
     <createSequence sequenceName="SEQ_PORTAL_PERMISSIONS_ID" startValue="1" />
   </changeSet>
 
+  <changeSet author="portal" id="1.0.0-30" dbms="postgresql">
+    <preConditions onFail="MARK_RAN">
+      <sqlCheck expectedResult="bytea">
+        SELECT DATA_TYPE FROM information_schema.columns where table_name = 'portal_windows' and column_name = 'customization'
+      </sqlCheck>
+    </preConditions>
+    <sql>
+      ALTER TABLE PORTAL_WINDOWS ADD COLUMN customization_oid oid;
+      UPDATE PORTAL_WINDOWS SET customization_oid = lo_from_bytea(0, customization);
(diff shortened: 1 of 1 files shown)


rdenarie:
cc @ahamdi

component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
```diff
@@ -329,4 +329,17 @@
...
\ No newline at end of file
+  <changeSet author="portal" id="1.0.0-30" dbms="postgresql">
+    <preConditions onFail="MARK_RAN">
+      <sqlCheck expectedResult="bytea">
+        SELECT DATA_TYPE FROM information_schema.columns where table_name = 'portal_windows' and column_name = 'customization'
+      </sqlCheck>
+    </preConditions>
+    <sql>
+      CREATE TABLE PORTAL_WINDOWS_lo AS SELECT id, title, icon,description,show_info_bar,show_app_state,show_app_mode,theme,width,height,properties,app_type,content_id, lo_from_bytea(0, customization) customization FROM PORTAL_WINDOWS;
+      DROP TABLE PORTAL_WINDOWS;
```
ahamdi: Could we just copy the content of the old column in a new column and then just drop the column instead of dropping the whole table, I think it will be safer and will take less time , WDYT ?

ahamdi (changes requested):

component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
rdenarie: You're, updated

ahamdi (approved):

PR #541, opened 2023-02-20
fix: when OpenID is not configured no warning should be displayed - EXO-62336

GouadriaHanen:
before this change, when OpenID is not configured, a malformed WellKnowConfigurationUrl error was displayed
since the WellKnowConfigurationUrl parameter does not have a default value, it is not empty "${exo.oauth.openid.wellKnownConfigurationUrl}", which leads to parameter checking that results in an error.
After this change, check if the provider is enabled or not to avoid chack in the disabled case.

### DIFF
--- a/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
+++ b/component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
@@ -428,24 +428,27 @@ public class OpenIdProcessorImpl implements OpenIdProcessor, Startable {
 
   @Override
   public void start() {
-    if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
-      log.info("wellKnownConfigurationUrl is not configured");
-      return;
-    }
-    try {
-      String wellKnownConfigurationContent = readUrl(new URL(this.wellKnownConfigurationUrl));
-      if (wellKnownConfigurationContent != null) {
-        JSONObject json = new JSONObject(wellKnownConfigurationContent);
-        this.authenticationURL = json.getString("authorization_endpoint");
-        this.accessTokenURL = json.getString("token_endpoint");
-        this.userInfoURL = json.getString("userinfo_endpoint");
-        this.issuer = json.getString("issuer");
-        this.remoteJwkSigningKeyResolver = new RemoteJwkSigningKeyResolver(this.wellKnownConfigurationUrl);
+    boolean openIdEnabled = Boolean.parseBoolean(System.getProperty("exo.oauth.openid.enabled"));
+    if (openIdEnabled) {
+      if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
+        log.error("wellKnownConfigurationUrl is not configured");
+        return;
       }
-    } catch (JSONException e) {
-      log.error("Unable to read webKnownUrl content : " + this.wellKnownConfigurationUrl);
-    } catch (MalformedURLException e) {
-      log.error("WellKnowConfigurationUrl malformed : url" + this.wellKnownConfigurationUrl);
+      try {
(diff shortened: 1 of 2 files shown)


component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -428,8 +428,12 @@ protected URL getUserInfoURL() throws IOException {
 
   @Override
   public void start() {
-    if (StringUtils.isBlank(this.wellKnownConfigurationUrl)) {
-      log.info("wellKnownConfigurationUrl is not configured");
+    boolean openIdEnabled = Boolean.parseBoolean(System.getProperty("exo.oauth.openid.enabled"));
```
rdenarie, line 431: What is the property is not defined ? Does Boolean.parseBoolean(null) works ?

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -428,8 +428,12 @@ protected URL getUserInfoURL() throws IOException {
...
+    if (!openIdEnabled) {
+      return;
+    }
```
```suggestion
    if (openIdEnabled) {
      
```
...
}

rdenarie (changes requested):

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
GouadriaHanen: it returns false

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
GouadriaHanen: done!

rdenarie:
Could you rebase, and resolve conflict ?
Thanks

GouadriaHanen:
> Could you rebase, and resolve conflict ? Thanks

done!

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -428,24 +428,27 @@ protected URL getUserInfoURL() throws IOException {
...
+        if (wellKnownConfigurationContent != null) {
+          JSONObject json = new JSONObject(wellKnownConfigurationContent);
+          this.authenticationURL = json.getString("authorization_endpoint");
+          this.accessTokenURL = json.getString("token_endpoint");
+          this.userInfoURL = json.getString("userinfo_endpoint");
+          this.issuer = json.getString("issuer");
+          this.remoteJwkSigningKeyResolver = new RemoteJwkSigningKeyResolver(this.wellKnownConfigurationUrl);
+        }
+      } catch (JSONException e) {
+        log.error("Unable to read webKnownUrl content : " + this.wellKnownConfigurationUrl);
```
```suggestion
        log.error("Unable to read webKnownUrl content : " + this.wellKnownConfigurationUrl,e);
```

component/web/oauth-common/src/main/java/org/gatein/security/oauth/openid/OpenIdProcessorImpl.java
```diff
@@ -428,24 +428,27 @@ protected URL getUserInfoURL() throws IOException {
...
+      } catch (MalformedURLException e) {
+        log.error("WellKnowConfigurationUrl malformed : url" + this.wellKnownConfigurationUrl);
```
```suggestion
        log.error("WellKnowConfigurationUrl malformed : url" + this.wellKnownConfigurationUrl,e);
```

rdenarie: Without this, the error is swallowed, and we not have the correct information in logs

rdenarie (changes requested):

rdenarie (approved):

PR #548, opened 2023-03-02
fix: TOP bar menu isn't displayed - EXO-62924

GouadriaHanen:
prior to this change, the top bar menu is not displayed and an error 500 is thrown since  it can not create a new sitekey with siteTypeName in uppercase
after this change, the siteTypeName is lowercased and the top bar menu is displayed

### DIFF
--- a/component/api/src/main/java/org/exoplatform/portal/mop/SiteKey.java
+++ b/component/api/src/main/java/org/exoplatform/portal/mop/SiteKey.java
@@ -63,11 +63,11 @@ public final class SiteKey implements Serializable {
 
     // This will be used for transition in usage from PortalKey and SiteKey
     public SiteKey(String type, String name) {
-        if (PortalConfig.PORTAL_TYPE.equals(type)) {
+        if (PortalConfig.PORTAL_TYPE.equalsIgnoreCase(type)) {
             this.type = SiteType.PORTAL;
-        } else if (PortalConfig.GROUP_TYPE.equals(type)) {
+        } else if (PortalConfig.GROUP_TYPE.equalsIgnoreCase(type)) {
             this.type = SiteType.GROUP;
-        } else if (PortalConfig.USER_TYPE.equals(type)) {
+        } else if (PortalConfig.USER_TYPE.equalsIgnoreCase(type)) {
             this.type = SiteType.USER;
         } else {
             throw new NullPointerException("No null name can be provided");
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
@@ -223,7 +223,7 @@ public class NavigationRest implements ResourceContainer {
       Collection<UserNode> nodes = new ArrayList<>();
 
       if (nodeId != null && siteName != null) {
-        SiteKey siteKey = new SiteKey(siteTypeName, siteName);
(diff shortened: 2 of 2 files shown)


rdenarie (commented):
Did we understand WHY the sitetypename is upper case on maintenance acceptance ? 
Why other instances are not impacted ?

Without answer to theses questions, we are not sure that this fix is the better thing to do.
cc @ahamdi

rdenarie:
In fact, the siteType is 'PORTAL' since few month, on Tribe, we can see 'PORTAL' also.
Why now, it is no more working with 'PORTAL' ?

rdenarie:
In fact, we have here a problem. 

Firstly the service v1/navigations is not correctly used in TopBarNavigationMenu. 
It is used by calling this url 
/portal/rest/v1/navigations/portal?siteType=PORTAL&siteName=dw....
but it should be called like this : 
/portal/rest/v1/navigations/PORTAL?siteName=dw....

This problem was fixed by Boubaker on develop here : 
https://github.com/Meeds-io/social/commit/30ce7e53b690a1ff17d3098b741ccb7f5b7d2df5#diff-5ca4ff1d7eff1f0e5dba8bb6ca0e57e756ba065318bffb35ece53df1c854857eR36

This problem leads to another problem : when creating a SiteKey, the siteType should be compared case incensitively. This modification was also done by Boubaker here : 
https://github.com/Meeds-io/gatein-portal/commit/2cc07cb10afa0287d1cd70b0709a4daab7ee7a0d#diff-f142ca2eb2e3d1e8a781ac7f489d82f3f076b5607c65341d65e66d60e7d20cc4R70
But this commit is still only on develop-meed and will arrive on develop in few days, we are reviewing the PR.
Once this PR will be merged, we will have the fix of this task on develop, without your fix.

So we can : 
- do nothing on develop, then the Boubaker PR will fix the problem. do nothing on stable/6.4.x as it is working, even if we are lucky that it is working
- Do nothing on develop, then the Boubaker PR will fix the problem. Then fix on 6.4.x NavigationService.js AND SiteKey.java. 
- Add the SiteKey fix on develop, but when the Boubaker's PR will arrive, we will have a conflict. AND fix on 6.4.x NavigationService.js AND SiteKey.java.

cc @ahamdi to discuss

ahamdi (dismissed):

rdenarie (changes requested):
This fix is not sufficient

rdenarie:
Do we put this fix on develop ? 
I think we should not as it will arrive via meeds mip merge.

cc @ahamdi

ahamdi (approved):

rdenarie:
@ahamdi I think we should not add this commit on develop, because it is coming in a short time with a meeds fix.

We should only make this on develop

GouadriaHanen:
> @ahamdi I think we should not add this commit on develop, because it is coming in a short time with a meeds fix.
> 
> We should only make this on develop

this fix will be only included in maintenance Ft and backported to 6.4.x

rdenarie:
ok, did you also create a PR to fix in social the TopBarMenu portlet ?

GouadriaHanen:
> ok, did you also create a PR to fix in social the TopBarMenu portlet ?

since the changes are available in the developt and maint FT, the PR will be created only for the backport after validating the changes on maint.

rdenarie:
ok, youre right, sorry

rdenarie (approved):

PR #550, opened 2023-03-08
fix: TOP bar menu isn't displayed - EXO-62924 -  Meeds-io/meeds#787

GouadriaHanen:
prior to this change, the top bar menu is not displayed and an error 500 is thrown since it can not create a new sitekey with siteTypeName in uppercase
after this change, the siteTypeName is  compared with ignoring cases and the top bar menu is displayed


ahamdi (changes requested):
We need a Meeds issue here

ahamdi (approved):

PR #561, opened 2023-03-29
feat: Move deprecated JS & CSS files - Meeds-io/MIPs#52 - MEED-1891

rdenarie:
Remove portlet_en.properties in translation.properties as it is removed.


rdenarie:
@boubaker it seems that you removed files portlet_*.properties in your last commit.

The line was not removed in translation.properties.
This commit remove this line as it break the sync crowdin job

boubaker (approved):

PR #564, opened 2023-04-03
feat: Implement delete node service - EXO-63071 - Meeds-io/meeds#51

IlhemEssaadi:
After this change, we will implement the delete node service.

### DIFF
--- a/component/api/src/main/java/org/exoplatform/portal/mop/service/NavigationService.java
+++ b/component/api/src/main/java/org/exoplatform/portal/mop/service/NavigationService.java
@@ -168,4 +168,10 @@ public interface NavigationService extends org.exoplatform.portal.mop.navigation
    */
   <N> void rebaseNode(NodeContext<N> context, Scope scope, NodeChangeListener<NodeContext<N>> listener);
 
+  /**
+   * Delete a navigation node with a given node id
+   *
+   * @param nodeId the node id to be deleted
+   */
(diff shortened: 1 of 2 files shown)


component/api/src/main/java/org/exoplatform/portal/mop/service/NavigationService.java
```diff
@@ -168,4 +168,10 @@ default <N> NodeContext<N> loadNodeById(NodeModel<N> model, // NOSONAR
    */
   <N> void rebaseNode(NodeContext<N> context, Scope scope, NodeChangeListener<NodeContext<N>> listener);
 
+  /**
+   * Delete node
```
azayati:
```suggestion
   * Delete a navigation node with a given node id
```

component/api/src/main/java/org/exoplatform/portal/mop/service/NavigationService.java
```diff
@@ -168,4 +168,10 @@ default <N> NodeContext<N> loadNodeById(NodeModel<N> model, // NOSONAR
...
+   *
+   * @param nodeId the node id to delete
```
azayati:
```suggestion
   * @param nodeId the node id to be deleted
```

azayati (changes requested):

azayati (approved):
LGTM

PR #565, opened 2023-04-03
feat: Update navigation REST service to retrieve only current site navigation - EXO-63564 - Meeds-io/meeds#51 

GouadriaHanen:
prior to this change, when retrieving navigation nodes for PORTAL with a specific siteName it returns all the nodes including nodes for extra site name (Global)
after this change, a new rest attribute is added to include the extra site in the result (includeGlobal)

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
@@ -7,11 +7,7 @@ import java.util.List;
 
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletRequest;
-import javax.ws.rs.GET;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
+import javax.ws.rs.*;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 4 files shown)


component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
```diff
@@ -140,13 +136,15 @@ public Response getSiteTypeNavigations(@Context HttpServletRequest request,
...
+                                         @Schema(defaultValue = "All possible values combined") @QueryParam("visibility") List<String> visibilityNames,
+                                         @Parameter(description = "if to include Global site in results in portal type case", required = false)
+                                         @DefaultValue("true")  @QueryParam("includeGlobal") boolean includeGlobal) {
     // this function return nodes and not navigations
     if (StringUtils.isBlank(siteTypeName)) {
       return Response.status(400).build();
     }
 
-    return getNavigations(request, siteTypeName, siteName, scopeName, nodeId, visibilityNames);
+    return getNavigations(request, siteTypeName, siteName, scopeName, nodeId, visibilityNames, includeGlobal);
```
azayati, line 147: we should set includeGlobal value to true when it is null

component/portal/src/main/java/org/exoplatform/portal/mop/user/UserPortal.java
```diff
@@ -115,7 +115,7 @@ default UserNode getNodeById(String userNodeId,
      * @param filterConfig an optional filter
      * @return a {@link Collection} of {@link UserNode}
      */
-    Collection<UserNode> getNodes(SiteType siteType, Scope scope, UserNodeFilterConfig filterConfig);
+    Collection<UserNode> getNodes(SiteType siteType, Scope scope, UserNodeFilterConfig filterConfig, boolean includeGlobal);
```
azayati, line 119: add includeGlobal param to the javadoc of this method

component/portal/src/main/java/org/exoplatform/portal/mop/user/UserPortalImpl.java
```diff
@@ -140,15 +140,16 @@ public List<UserNavigation> getNavigations() {
...
 
     Collection<UserNode> resultUserNodes = new ArrayList<>();
     Set<String> addedUserNodesURI = new HashSet<>();
     for (UserNavigation userNavigation : getNavigations()) {
       SiteKey siteKey = userNavigation.getKey();
       if (siteKey.getType() != siteType
           || (siteType == SiteType.GROUP && siteKey.getName().startsWith(SPACES_SITE_TYPE_PREFIX))
-          || (siteType == SiteType.SPACE && !siteKey.getName().startsWith(SPACES_SITE_TYPE_PREFIX))) {
+          || (siteType == SiteType.SPACE && !siteKey.getName().startsWith(SPACES_SITE_TYPE_PREFIX))
+          || (!includeGlobal && siteKey.getName().equalsIgnoreCase("global"))) {
```
azayati: WDYT if we modify getNavigations and loadUserNavigation methods instead by changing the condition:

```java
if (siteKey.getType() == SiteType.PORTAL
          && StringUtils.isNotBlank(service.getGlobalPortal())
          && !StringUtils.equals(service.getGlobalPortal(), siteKey.getName())) {
        // Add global navigation at the end
        loadUserNavigation(new SiteKey(SiteType.PORTAL, service.getGlobalPortal()));
 }
```

azayati (changes requested):

component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
GouadriaHanen: it is boolean could not be null  and already has a default value "true"

component/portal/src/main/java/org/exoplatform/portal/mop/user/UserPortalImpl.java
GouadriaHanen: in "PORTAL" case, we need to modify getNodes method

azayati (approved):
LGTM

PR #580, opened 2023-04-14
feat: Change get membership types endpoint roles allowed to users - EXO-63142 - Meeds-io/meeds#51

GouadriaHanen:
this change is to change get membership types endpoint roles allowed to users


azayati (approved):
LGTM

PR #581, opened 2023-04-14
feat: Implement createNode API service - EXO-63243 - Meeds-io/meeds#51

IlhemEssaadi:
After this change, we will implement the add node service.


component/api/src/main/java/org/exoplatform/portal/mop/service/NavigationService.java
```diff
@@ -194,4 +195,15 @@ default <N> NodeContext<N> loadNodeById(NodeModel<N> model, // NOSONAR
...
    */
   public void moveNode(Long targetId, Long fromId, Long toId, Long previousId);
+
+  /**
+   * Create a navigation node
+   *
+   * @param parentId the parent node id
+   * @param previousId the previous node id
+   * @param name node name
+   * @param state
```
azayati:
```suggestion
   * @param state node state
```

azayati (changes requested):

azayati (approved):

PR #582, opened 2023-04-17
feat: Implement createNode API service - EXO-63243 - Meeds-io/meeds#51

IlhemEssaadi:
After this change, we will implement the add node service.


azayati (approved):
LGTM

PR #583, opened 2023-04-17
feat: Expand node page extra details on navigation rest service results - EXO-63142_EXO-63143 - Meeds-io/meeds#51

GouadriaHanen:
this change is added to extend extra attributes for navigation nodes' DTO model


azayati (approved):
LGTM

PR #588, opened 2023-04-24
feat: Implement temporalCheck API param to getNavigations rest API - EXO-63240 - Meeds-io/MIPs#51

IlhemEssaadi:
Prior to this change, when retrieving navigation nodes, a temporal check for scheduled nodes is always applied in order to retrieve only those respecting the start and end scheduling dates interval. 
For some cases, we need to always retrieve scheduled nodes regardless start and end scheduling dates, for that we have added the temporalCheck boolean param for getNavigations rest API to specify if the temporal check should be applied or not.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
@@ -56,7 +56,7 @@ public class NavigationRest implements ResourceContainer {
 
   private static final Visibility[]         DEFAULT_VISIBILITIES = Visibility.values();
 
-  private static final UserNodeFilterConfig USER_FILTER_CONFIG   = getUserFilterConfig(DEFAULT_VISIBILITIES);
+  private static final UserNodeFilterConfig USER_FILTER_CONFIG   = getUserFilterConfig(DEFAULT_VISIBILITIES, false);
 
   private UserPortalConfigService           portalConfigService;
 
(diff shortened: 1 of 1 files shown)


component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
```diff
@@ -159,13 +159,17 @@ public Response getSiteTypeNavigations(@Context
                                          boolean includeGlobal,
                                          @Parameter(description = "to include extra node page details in results")
                                          @QueryParam("expandPageDetails")
-                                         boolean expandPageDetails) {
+                                         boolean expandPageDetails,
+                                         @Parameter(description = "to check start and end time of node publication")
```
azayati:
```suggestion
                                         @Parameter(description = "to check the navigation nodes scheduling start and end dates")
```

component/portal/src/main/java/org/exoplatform/portal/mop/rest/NavigationRest.java
```diff
@@ -227,8 +232,8 @@ private Response getNavigations(HttpServletRequest request, // NOSONAR
     }
 
     UserNodeFilterConfig userFilterConfig = USER_FILTER_CONFIG;
-    if (visibilities.length > 0) {
-      userFilterConfig = getUserFilterConfig(visibilities);
+    if (visibilities.length > 0 || temporalCheck) {
```
azayati, line 235: can we add a unit test for that ?

azayati (changes requested):

azayati (changes requested):
see review

PR #593, opened 2023-04-28
feat: Implement temporalCheck API param to getNavigations rest API - EXO-63240 - Meeds-io/MIPs#51

IlhemEssaadi:
Prior to this change, when retrieving navigation nodes, a temporal check for scheduled nodes is always applied in order to retrieve only those respecting the start and end scheduling dates interval.
For some cases, we need to always retrieve scheduled nodes regardless start and end scheduling dates, for that we have added the temporalCheck boolean param for getNavigations rest API to specify if the temporal check should be applied or not.


azayati (approved):
LGTM

PR #601, opened 2023-05-18
feat: Implement navigation node description API service - EXO-63408 - Meeds-io/MIPs#51

IlhemEssaadi:
This change is to implement navigation node description API service.


azayati (approved):
LGTM

PR #610, opened 2023-06-02
fix: Clean insert left/right dropdown from column container - EXO-64611 - Meeds-io/MIPs#51

IlhemEssaadi:
Prior to this fix, bts_dropdown requirejs is removed from the template since bootstrap is nomore needed in Meeds which makes the insert right/left action not working since it is based on the bts_dropdown js. After this commit, since this action can be performed by interface when add or edit a page which is not available on meed, we have cleaned the template by removing this useless action.

### DIFF
--- a/web/portal/src/main/webapp/groovy/portal/webui/container/UISimpleColumnContainer.gtmpl
+++ b/web/portal/src/main/webapp/groovy/portal/webui/container/UISimpleColumnContainer.gtmpl
@@ -6,6 +6,8 @@
   
   def rcontext = _ctx.getRequestContext();
   ResourceBundle res = rcontext.getApplicationResourceBundle();
+  def requirejs = rcontext.getJavascriptManager().getRequireJS();
+  requirejs.require("SHARED/bts_dropdown","bts_dropdown");
 
   UIPortalApplication uiPortalApp = rcontext.getUIApplication();
   boolean hasPermission = uicomponent.hasPermission();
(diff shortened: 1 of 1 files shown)


web/portal/src/main/webapp/groovy/portal/webui/container/UISimpleColumnContainer.gtmpl
```diff
@@ -110,7 +112,7 @@
                       </div>
                       <a href="javascript:void(0);" onclick="<%=uicomponent.event("EditContainer")%>" rel="tooltip" data-placement="bottom" title="<%=_ctx.appRes("UIColumnContainer.tooltip.editContainer")%>"><i class="uiIconEdit uiIconWhite"></i></a>
                       <a href="javascript:void(0);" onclick="<%=uicomponent.event("DeleteComponent")%>" rel="tooltip" data-placement="bottom" title="<%=_ctx.appRes("UIColumnContainer.tooltip.closeContainer")%>"><i class="uiIconTrash uiIconWhite"></i></a>
-                    <%}%>                   
+                    <%}%
```
azayati:
```suggestion
                    <%}%>
```

azayati (changes requested):

web/portal/src/main/webapp/groovy/portal/webui/container/UISimpleColumnContainer.gtmpl
```diff
@@ -6,6 +6,8 @@
   
   def rcontext = _ctx.getRequestContext();
   ResourceBundle res = rcontext.getApplicationResourceBundle();
+  def requirejs = rcontext.getJavascriptManager().getRequireJS();
+  requirejs.require("SHARED/bts_dropdown","bts_dropdown");
```
azayati: > We may need to not use bottstrap JS anymore if possible. If it's used by a WebUI or Juzu portlet, please add it as dependency of the portlet itself. This container will not work in Meeds context at all and it will be buggy since the bootstrap JS has been deleted from Meeds files (it's added in eXo side only, please read more the details in [Meeds-io/MIPs#52](https://github.com/Meeds-io/MIPs/issues/52) and its related PRs descriptions).


sorry @boubaker your comment is deleted by mistake, in our case the UISimpleColumnContainer.gtmpl is not used in a specific portlet, it is can be added by interface in any page or site layout and we need the dropdown to insert right/left containers. Should we overwrite it in eXo context ?

azayati (changes requested):

azayati (approved):
LGTM

PR #618, opened 2023-07-03
feat: Refactor findSites method to ignore empty sites - EXO-65013 - Meeds-io/MIPs#68

GouadriaHanen:
Prior to this change, sites were filtered using the includeEmptySites filter only when pagination is needed.
In our case, as we have more than one site, the pagination is set to null, so all sites are included, and the includeEmptySites filter is not taken into account.
After this change, in all cases (with/without pagination) the sites will be filtered and empty sites will not loaded.


azayati (approved):
LGTM

azayati (changes requested):
Please add UT

azayati (approved):
LGTM

PR #624, opened 2023-07-07
feat: Change the character set for description column of PORTAL_SITES table to utf8mb4 in order to support emojis insertion - EXO-65345 - Meeds-io/MIPs#68

IlhemEssaadi:
Prior to this change, the default character set for the description column in the PORTAL_SITE table is utf8, so we can't add text containing emojis. After this change, we replace the column's character set with utf8mb4 so that the description column can support emojis.

### DIFF
--- a/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
+++ b/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
@@ -367,4 +367,9 @@
       <column name="UPDATED_DATE" type="BIGINT" defaultValueNumeric="0" />
     </addColumn>
   </changeSet>
+  <changeSet author="portal" id="1.0.0-34" dbms="mysql">
+    <sql>
+      ALTER TABLE PORTAL_SITES MODIFY COLUMN DESCRIPTION longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+    </sql>
+  </changeSet>
 </databaseChangeLog>


component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
```diff
@@ -367,4 +367,9 @@
       <column name="UPDATED_DATE" type="BIGINT" defaultValueNumeric="0" />
     </addColumn>
   </changeSet>
+  <changeSet author="portal" id="1.0.0-34" dbms="mysql">
+    <sql>
+      ALTER TABLE PORTAL_SITES MODIFY COLUMN DESCRIPTION longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
ahamdi, line 372: Wouldn't it be enough to create database with UTF charset without putting it here in the liquibase script ?

azayati: @ahamdi The default charset of PORTAL_SITES table is set on creation explicitly to CHARSET=UTF8 COLLATE utf8_general_ci. https://github.com/Meeds-io/gatein-portal/blob/a488e09656036b801e6490f1a264c2b11f509217/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml#L81
The description field will inherit the same charset of its corresponding table.
So even if we create the database with **utf8mb4 charset**, the charset of the PORTAL_SITES table is not changed and remains **utf8_general_ci**, so IMO we need to change the charset of the description column via liquibase as it is done for other tables cases.
Same examples for:

- tasks description field https://github.com/Meeds-io/task/blob/develop/services/src/main/resources/db/changelog/task.db.changelog-1.0.0.xml#L449
- activities body field https://github.com/Meeds-io/social/blob/3c2d18e8748804f9d5a5a9560794e08e835bad35/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml#L562

azayati (approved):
LGTM

component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
ahamdi: @azayati You need to change the charset to the correct one for the whole tables. can you change the collate in [gatein-portal/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml](https://github.com/Meeds-io/gatein-portal/blob/a488e09656036b801e6490f1a264c2b11f509217/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml#L81)

rdenarie:
Why do we need a different charset ? why utf-8 is not sufficient ?
Can you explain it at least in the commit message which is currently empty ?

Thanks

component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
azayati: @ahamdi you mean that we should change the charset for the whole PORTAL_SITES table ?
IMO since only the description field is considered, why we don't change only the charset of the field itself ?

azayati (changes requested):
@IlhemEssaadi you can apply the needed charset only for the description field, and according to that update the PR description
We will discuss with ITOP the possibility to apply the charset to the whole db

exo-swf:

azayati (approved):
LGTM

PR #628, opened 2023-07-12
feat: Implement Argon2PasswordEncoder - EXO-59676  - Meeds-io/MIPs#69

hakermi:
Prior to this change, current password hashing algorithm is SHA-256 with a salt of 64 bits generated by a random algorithm (SHA1PRNG). Our need to a stronger hashing mechanism that will complexify and slow attacks has lead us to Argon2 which is more pertinent for a password hashing algorithm as recommended by ANSSI.
In This PR we implement a new password encoder that uses argon2 to encode and hash the passwords

### DIFF
--- a/component/web/security/pom.xml
+++ b/component/web/security/pom.xml
@@ -61,6 +61,10 @@
       <groupId>org.gatein.captcha</groupId>
       <artifactId>simplecaptcha</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk15on</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/component/web/security/src/main/java/org/exoplatform/web/security/hash/Argon2IdPasswordEncoder.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/security/hash/Argon2IdPasswordEncoder.java
@@ -1,0 +1,158 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ * Copyright (C) 2023 Meeds Association
+ * contact@meeds.io
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+*/
+package org.exoplatform.web.security.hash;
+
+import org.apache.commons.codec.binary.Hex;
+import org.bouncycastle.crypto.generators.Argon2BytesGenerator;
+import org.bouncycastle.crypto.params.Argon2Parameters;
+import org.exoplatform.container.PortalContainer;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
+import org.exoplatform.web.security.codec.AbstractCodec;
+import org.exoplatform.web.security.codec.CodecInitializer;
+import org.exoplatform.web.security.security.TokenServiceInitializationException;
+import org.picketlink.idm.api.Attribute;
+import org.picketlink.idm.api.AttributesManager;
+import org.picketlink.idm.api.CredentialEncoder;
+import org.picketlink.idm.api.SecureRandomProvider;
+import org.picketlink.idm.common.exception.IdentityException;
+import org.picketlink.idm.impl.credential.HashingEncoder;
+
+import java.nio.charset.StandardCharsets;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class Argon2IdPasswordEncoder extends HashingEncoder {
+
+  private static final Log        LOG                                               =
+                                      ExoLogger.getLogger(Argon2IdPasswordEncoder.class);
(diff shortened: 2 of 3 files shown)


component/identity/src/test/java/org/gatein/portal/idm/impl/encoder/Argon2PasswordEncoderTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class Argon2PasswordEncoderTest {
+
+  private Argon2PasswordEncoder argon2PasswordEncoder;
+
+  @Before
+  public void setUp() throws Exception {
+    System.setProperty("credentialEncoder.secureRandomAlgorithm", "SHA1PRNG");
```
rdenarie: Why do we need this property in the test ?

component/identity/src/main/java/org/gatein/portal/idm/impl/encoder/Argon2PasswordEncoder.java
```diff
@@ -0,0 +1,80 @@
...
+  public String encodeCredential(String username, String rawPassword) {
+    Argon2Parameters.Builder builder =
+                                     new Argon2Parameters.Builder(Argon2Parameters.ARGON2_id).withIterations(ARGON2_HASHING_ITERATIONS)
+                                                                                             .withMemoryAsKB(ARGON2_MEMORY_LIMIT)
+                                                                                             .withParallelism(ARGON2_PARALLEL_EXECUTIONS)
+                                                                                             .withSalt(getStoredSalt(username));
+    Argon2BytesGenerator generator = new Argon2BytesGenerator();
+    generator.init(builder.build());
+    byte[] hash = new byte[ARGON2_HASH_LENGTH];
+    generator.generateBytes(rawPassword.getBytes(StandardCharsets.UTF_8), hash, 0, hash.length);
```
rdenarie: In the specification, we speak about a symetrical encryption at the end of the encode process, using CodecInitializer.

Could you add it ?

rdenarie (changes requested):

component/identity/src/test/java/org/gatein/portal/idm/impl/encoder/Argon2PasswordEncoderTest.java
hakermi: It is needed in the initialize in the inherited `afterInitialize() `function

component/identity/src/main/java/org/gatein/portal/idm/impl/encoder/Argon2PasswordEncoder.java
hakermi: Done

component/web/security/src/main/java/org/exoplatform/web/security/hash/Argon2PasswordEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+import org.exoplatform.web.security.security.TokenServiceInitializationException;
+import org.picketlink.idm.api.Attribute;
+import org.picketlink.idm.api.AttributesManager;
+import org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder;
+
+import java.nio.charset.StandardCharsets;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class Argon2PasswordEncoder extends DatabaseReadingSaltEncoder {
```
rdenarie:
```suggestion
public class Argon2PasswordEncoder extends HasingEncoder {
```

We dont want to extends DatabaseReadingSaltEncoder, as it use an algo from the JVM. 
We want a new encoder, so we should extends HashingEncoder only

component/web/security/src/main/java/org/exoplatform/web/security/hash/Argon2PasswordEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+    return salt;
+  }
+
+  private static AbstractCodec getCodec() throws TokenServiceInitializationException {
+    if (codecInitializer != null) {
+      return codecInitializer.getCodec();
+    }
+    PortalContainer container = PortalContainer.getInstance();
+    codecInitializer = container.getComponentInstanceOfType(CodecInitializer.class);
+    return codecInitializer.getCodec();
```
rdenarie: Could we make it more readable ?
```suggestion
    if (codecInitializer == null) {
      PortalContainer container = PortalContainer.getInstance();
      this.codecInitializer = container.getComponentInstanceOfType(CodecInitializer.class);
    }
   
    return codecInitializer.getCodec();
```

rdenarie (changes requested):

component/web/security/src/main/java/org/exoplatform/web/security/hash/Argon2PasswordEncoder.java
hakermi: I didn't understand what you mean of using an algo from the JVM, i extended `DatabaseReadingSaltEncoder` which extends the `HashingEncoder` to not rewrite the `afterInitialize()` and the `getSecureRandomInstance(`) functions
and to still be able to set the secureRandomAlgoithm by server prop.
because if i extend directly the HashingEncoder then i will copy past the two functions from DatabaseReadingSaltEncoder so it doesn't make any sense

rdenarie: It for class organization
The "old" encoder is DatabaseReadingEncoder.
We create a new one => so new class.
The common class between the two is HashingEncoder

As Argon2idEncoder have nothing common with DatabaseReadingEncoder except salt usage, I dont want that the new encoder extends the old one.

rdenarie:
```suggestion
public class Argon2IdPasswordEncoder extends DatabaseReadingSaltEncoder {
```

rdenarie (changes requested):

rdenarie (approved):

PR #629, opened 2023-07-13
feat: Change the character set for description column of PORTAL_SITES table to utf8mb4 in order to support emojis insertion - EXO-65345 - Meeds-io/MIPs#68

IlhemEssaadi:
Prior to this change, the default character set for the description column in the PORTAL_SITE table is utf8, so we can't add text containing emojis. After this change, we replace the column's character set with utf8mb4 so that the description column can support emojis.


azayati (approved):
LGTM

PR #630, opened 2023-07-18
feat: Update the default password hashing algorithm - EXO-65150  - Meeds-io/MIPs#69

hakermi:
Prior to this change, Old created passwords are hashed by the new encoder algorithm and need an upgrade mechanism  to re-hash theses password without asking users to create new ones.
This PR is a part of the upgrade mechanism, will check the password slat attribute and validate it by the proper algorithm, if it's hashed by the old algorithm then it will be updated by the new one.

### DIFF
--- a/component/identity/src/main/resources/picketlink-idm-config.xml
+++ b/component/identity/src/main/resources/picketlink-idm-config.xml
@@ -37,11 +37,15 @@
         </option>
         <option>
           <name>credentialEncoder.class</name>
-          <value>org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder</value>
+          <value>${exo.plidm.password.class:org.exoplatform.web.security.hash.Argon2IdPasswordEncoder}</value>
         </option>
         <option>
           <name>credentialEncoder.hashAlgorithm</name>
-          <value>SHA-256</value>
+          <value>${exo.plidm.password.hash:SHA-256}</value>
+        </option>
+        <option>
+          <name>credentialEncoder.secureRandomAlgorithm</name>
+          <value>${exo.plidm.password.secureRandomAlgorithm:SHA1PRNG}</value>
         </option>
       </options>
     </realm>
--- a/component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
+++ b/component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
@@ -30,6 +30,24 @@
       <identity-type-mappings>
         <user-mapping>USER</user-mapping>
       </identity-type-mappings>
+      <options>
+        <option>
+          <name>template</name>
+          <value>true</value>
+        </option>
+        <option>
+          <name>credentialEncoder.class</name>
+          <value>org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder</value>
+        </option>
+        <option>
+          <name>credentialEncoder.hashAlgorithm</name>
+          <value>SHA-256</value>
+        </option>
+        <option>
+          <name>credentialEncoder.secureRandomAlgorithm</name>
+          <value>SHA1PRNG</value>
+        </option>
+      </options>
     </realm>
   </realms>
   <repositories>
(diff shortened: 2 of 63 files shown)


component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManagerImpl.java
```diff
@@ -0,0 +1,54 @@
...
+                                                                       "org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder");
+
+  private static PicketLinkIDMService idmService;
+
+  public ExtendedAttributeManagerImpl(IdentitySessionImpl session) {
+    super(session);
+  }
+
+  public boolean validatePassword(User user, String password) throws IdentityException {
+    DatabaseReadingSaltEncoder oldCredentialEncoder = new DatabaseReadingSaltEncoder();
```
rdenarie: The pseudo code in the tech study is :

> check if user have attribute passwordSalt128
> if yes : use the current way to check password
> it will get the default configured PasswordEncoder (Argon2IdPasswordEncoder)
> and use it
> 
> if no, validateCredential by using the old passwordEncoder :
> DatabaseReadingSaltEncoder
> it will automatically use the old configured algo to validate the password
> (SHA256 or MD5 in function of the configuration)
> If validation is ok, call updatePassword to compute and save a hash with the new algo. During passwordUpdate, we need to ensure to remove passwordSalt

It has the advantage to not initialize oldEncoder if not needed. 
In addition, I wish that, if possible, check the password with the correct encoder : I dont want to test with the old one, and if not work, test the new one, because this solution implies to make 2 verification, it is not performant.

By checking the attribute passwordSalt, we have the possibility to verify if the new algo must be used, or the old one.

rdenarie (changes requested):

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManagerImpl.java
hakermi: done

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
```diff
@@ -30,6 +30,24 @@
...
         <user-mapping>USER</user-mapping>
       </identity-type-mappings>
+      <options>
+        <option>
+          <name>template</name>
+          <value>true</value>
+        </option>
+        <option>
+          <name>credentialEncoder.class</name>
+          <value>org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder</value>
```
rdenarie, line 40: Should we keep the old encoder here ?

component/identity/src/main/resources/picketlink-idm-config.xml
```diff
@@ -37,7 +37,7 @@
         </option>
         <option>
           <name>credentialEncoder.class</name>
-          <value>org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder</value>
+          <value>org.exoplatform.web.security.hash.Argon2IdPasswordEncoder</value>
```
rdenarie: should we use the property to define the encoder ?

rdenarie (changes requested):

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
hakermi: the new one will cause an issue in the integration tests of userDaoImpl,  because the web.security dependencies is not available, using the old one will will create a user with the old encoder and then due to the integration test, it will update the user password by the new encoder

component/identity/src/main/resources/picketlink-idm-config.xml
hakermi: done

rdenarie (approved):

PR #635, opened 2023-07-27
feat: Update password validation in authentication phase regarding the new hash algorithm upgrade - EXO-65151 - Meeds-io/MIPs#69

hakermi:
Prior to this change, As a final step in password hash upgrade old authentication process need to be updated to take into consideration the modification in passwords hashing.
This PR updates the authentication process to validate passwords by the new hash algorithm including the old hashed passwords after an UP that has updated the old hashing

### DIFF
--- a/component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
+++ b/component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
@@ -18,26 +18,32 @@ package org.gatein.portal.idm.impl.store.attribute;
 
 import org.exoplatform.container.ExoContainer;
 import org.exoplatform.container.ExoContainerContext;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
 import org.exoplatform.services.organization.idm.PicketLinkIDMService;
 import org.picketlink.idm.api.Attribute;
+import org.picketlink.idm.api.CredentialEncoder;
 import org.picketlink.idm.api.User;
 import org.picketlink.idm.common.exception.IdentityException;
 import org.picketlink.idm.impl.api.PasswordCredential;
 import org.picketlink.idm.impl.api.session.IdentitySessionImpl;
 import org.picketlink.idm.impl.api.session.managers.AttributesManagerImpl;
-import org.picketlink.idm.impl.credential.DatabaseReadingSaltEncoder;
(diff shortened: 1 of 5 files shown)


component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
```diff
@@ -44,24 +45,33 @@ public ExtendedAttributeManager(IdentitySessionImpl session) {
...
     Attribute salt128 = getAttribute(user.getKey(), PASSWORD_SALT_USER_ATTRIBUTE);
-    if (salt128 != null) {
-      return super.validatePassword(user, password);
-    } else {
-      DatabaseReadingSaltEncoder oldCredentialEncoder = new DatabaseReadingSaltEncoder();
-      oldCredentialEncoder.setIdentitySession(getIdentitySession());
-      oldCredentialEncoder.initialize(getCredentialEncoderProps(),
-                                      getIdmService().getIdentityConfiguration().getIdentityConfigurationRegistry());
+    if (salt != null && salt128 != null) {
+      return super.validatePassword(user,
```
rdenarie:
```suggestion
      //password has been updated by upgradePlugin, and user was not connected since
      //passwordHash was created like this passwordHash = newEncoder(oldEncoder(password))
      return super.validatePassword(user,
```

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
```diff
@@ -44,24 +45,33 @@ public ExtendedAttributeManager(IdentitySessionImpl session) {
...
+                                    (String) new PasswordCredential(password,
+                                                                    getOldCredentialEncoder(),
+                                                                    user.getKey()).getEncodedValue());
+    } else if (salt != null) {
```
rdenarie:
```suggestion
    } else if (salt != null) {
       //user didnt reconnect after new passwordEncoder modification, but upgrade plugin was not executed for him, his passwordHash is still oldEncoder(password)
```

rdenarie (changes requested):

rdenarie (approved):

rdenarie (approved):

rdenarie (approved):

exo-swf:

rdenarie (approved):

PR #639, opened 2023-08-08
feat: Update password validation in authentication phase regarding the new hash algorithm upgrade - EXO-65151 - Meeds-io/MIPs#69

hakermi:
Prior to this change, As a final step in password hash upgrade old authentication process need to be updated to take into consideration the modification in passwords hashing.
This PR updates the authentication process to validate passwords by the new hash algorithm including the old hashed passwords after an UP that has updated the old hashing


rdenarie (changes requested):
please make this pr on develop-exo

hakermi:
> please make this pr on develop-exo

Done

rdenarie (approved):

PR #640, opened 2023-08-09
fix: Improve security for rememberme feature - EXO-59675 - meeds-io/MIPs#69

rdenarie:
Before this fix, user password was encoded in the rememberme token, with a reversible function This commit update the rememberme to not use the user password in the rememberme token


hakermi (approved):

hakermi (commented):
to fix sonar issues

rdenarie:
> to fix sonar issues

Done.

PR #642, opened 2023-08-10
Merge mip 69

### LINKED ISSUE #69: [Technical Improvement] Update Password Encoder Algorithm
## Rationale
Our password encoder algorithm is outdated and not secure (SHA-256)
We need to update to a more secure one

In addition, the usage of RememberMe token is not secure.

## 1. Functional Requirements
No functionnal modification 

### Security 
We want to use the Argon2id algorithm to hash passwords with parameters 
·	int opsLimit = 2; // 2 iterations
·	int memLimit = 32768; // 32 MB
·	int parallelism = 1; // 1 parallel execution
·	int outputLength = 32; // password hash length
The password Salt has to be 128 bits (16 bytes) of length.

For RememberMe Token, we want to generate it with the Argon2id algorithm with same parameters as password

### Configurability 
As Argon2 parameters could not be changed without reseting all password, we not make it configurable

### Upgradability
The migration strategy have 3 options : 
- keep hash with old algo. When user logs in, update the password hash with the new algo
- Invalidate passwords, so thats users have to update password, which will be hashed with new algo. Not applicable in our case as we cannot "invalidate" password
- Update SHA-256 hashed passwords stored in database. Database stores as password hash oldAlgo(password). Upgrade plugin will update theses hashes to have newAlgo(oldAlgo(password)). So that, all password are hashed with the new algo. The login processus must be able to validate password in both options.

We will use the most secure strategy based on data upgrade : the solution 3

### Existing Features
No modification

### Feature Flags
No Feature flag

### Documentation 
Technical documentation to update

## 4. Software Architecture
Store password process : 
image

Store Password Sequence Diagramm
image

Only the red part is to create

Validate Password process
image

Validate Password pseudo code in AttributeManager.validatePassword (afeter data upgrade execution)
```
Input : username, password

If (not user have attribute passwordSalt128 && user have attribute passwordSalt) {
   // user not connected && not migrated by data upgrade
   Check password with old algo (oldAlgo(password) == hashInDatabase ?)
   If validation is ok, call updatePassword to compute and save a hash with the new algo. During passwordUpdate, we need to ensure to remove passwordSalt attribute

} else if(user have attribute passwordSalt128 && user have attribute passwordSalt) (
   // user not connected && migrated by data upgrade
   Hash = Compute the hash of the password by using the old passwordEncoder  DatabaseReadingSaltEncoder. (oldAlgo(password))
   Then check hash with newAlgo (newAlgo(oldAlgo(password)) == hashInDatabase ?)
   If validation is ok, call updatePassword to compute and save a hash with the new algo. During passwordUpdate, we need to ensure to remove passwordSalt attribute

} else {
  //user connected
  use the current way to check password it will get the default configured PasswordEncoder and use it
      return this.getRepository().validateCredential(this.getInvocationContext(), this.createIdentityObject(user), new PasswordCredential(hash, this.getCredentialEncoder(), user.getKey()));
}

```

Remember me generation : 
image

RememberMe Validation
image

Once remember me token are generated with new algo, we need an AuthenticatorPlugin to authenticate the user.

### Access 
   - GUI 
   - API 

### Services & processing
   - Ex: Scheduled Jobs

### Data and persistence
   - DataUpgrade 

```
for all user without attribute “passwordSalt128” :
   Get password hash in database
   Generate a passwordSalt128 and store it as attribute for the user
   Compute newAlgo(hash) and store it as new hash for the user
```
## 5. Annexes
   - Impacts 
   - References

rdenarie:


component/identity/src/main/java/org/exoplatform/services/organization/idm/PicketLinkIDMServiceImpl.java
```diff
@@ -210,6 +212,16 @@ public IdentitySession getIdentitySession(String realm) throws Exception {
...
     }
 
+    @Override
+    public IdentityConfiguration getIdentityConfiguration() {
+        return identityConfiguration;
+    }
+
+    @Override
+    public ExtendedAttributeManager getExtendedAttributeManager() throws Exception {
+      return new ExtendedAttributeManager((IdentitySessionImpl) getIdentitySession());
```
boubaker: Shouldn't we avoid building anew object each time this getter is invoked ?

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
```diff
@@ -0,0 +1,131 @@
...
+  public static final String          PASSWORD_SALT_USER_ATTRIBUTE     = "passwordSalt128";
+
+  private static final String         DEFAULT_ENCODER                  =
+                                                      "org.exoplatform.web.security.hash.Argon2IdPasswordEncoder";
+
+  private static final String         OLD_ENCODER_CLASS_PROPERTY       = "oldCredentialEncoder.class";
+
+  private static PicketLinkIDMService idmService;
+
+  public ExtendedAttributeManager(IdentitySessionImpl session) {
```
boubaker, line 50: Can we make this as a service and move `session` parameter into `validatePassword` method as parameter ?

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
```diff
@@ -30,6 +30,24 @@
...
+        </option>
+        <option>
+          <name>credentialEncoder.hashAlgorithm</name>
+          <value>SHA-256</value>
+        </option>
+        <option>
+          <name>credentialEncoder.secureRandomAlgorithm</name>
+          <value>SHA1PRNG</value>
+        </option>
+      </options>
```
boubaker, line 50: Should move this into `supported-identity-object-types`/`<identity-object-type>`/`<options>` of `HibernateStore`.
In fact, when declaring `HibernateStore` and other stores, like `LDAP`, this option will have to be defined for Hibernate only. BTW, please update other PLIDM config files in the same folder, section `HibernateStore`.

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+      } else {
+        log.debug("No username has been committed. RemembermeLoginModule will be ignored.");
+      }
+
+      return true;
+    } catch (final Exception e) {
+      log.warn(e.getMessage());
+      throw new LoginException(e.getMessage());
```
boubaker: either log or throw exception, never both

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
```diff
@@ -0,0 +1,88 @@
...
+import org.exoplatform.web.security.security.CookieTokenService;
+import org.gatein.sso.agent.tomcat.ServletAccess;
+
+public class RememberMeLoginModule extends TomcatLoginModule {
+
+  private static final Log   log                = ExoLogger.getLogger(RememberMeLoginModule.class);
+
+  @Override
+  public boolean login() throws LoginException {
+    log.debug("In login of RemembermeLoginModule.");
```
boubaker: useless ?

```suggestion
```

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
```diff
@@ -0,0 +1,88 @@
...
+    try {
+
+      HttpServletRequest servletRequest = ServletAccess.getRequest();
+      if (servletRequest == null) {
+        LOG.warn("HttpServletRequest is null. RemembermeLoginModule will be ignored.");
+        return false;
+      }
+      String token = LoginUtils.getRememberMeTokenCookie(servletRequest);
+
+      if (token!=null) {
```
boubaker: formatting (please format the whole file)

```suggestion
      if (token != null) {
```

boubaker: please format the whole file

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (C) 2013 eXo Platform SAS.
```
boubaker: Wrong License Header
```suggestion
```

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
```diff
@@ -0,0 +1,131 @@
...
+    super(session);
+  }
+
+  @Override
+  public boolean validatePassword(User user, String password) throws IdentityException {
+    Attribute salt128 = getAttribute(user.getKey(), PASSWORD_SALT_USER_ATTRIBUTE);
```
boubaker: We don't need to add the length of salt in variable names for better future code upgradability

```suggestion
    Attribute salt = getAttribute(user.getKey(), PASSWORD_SALT_USER_ATTRIBUTE);
```

component/web/security/src/main/java/org/exoplatform/web/security/security/CookieTokenService.java
```diff
@@ -135,34 +77,29 @@ public void start() {
...
-
-            /* the symmetric encryption happens here */
-            String encryptedPassword = codec.encode(credentials.getPassword());
-            Credentials encodedCredentials = new Credentials(credentials.getUsername(), encryptedPassword);
-
+            cookieTokenString=selector+SEPARATOR_CHAR+validator;
             try {
-                this.tokenStore.saveToken(new GateInTokenStore.TokenData(id, hashedRandomString, encodedCredentials, new Date(expirationTimeMillis), type));
-//                tokenContainer.saveToken(context.getSession(), id, hashedRandomString, encodedCredentials, new Date(expirationTimeMillis, tokenType));
+                this.tokenStore.saveToken(new GateInTokenStore.TokenData(selector, hashedRandomString,
```
boubaker, line 101: A security layer has been removed here when removing the usage of `CodecInitializer` to encrypt part of the content. In fact, we 'll need to preserve the usage of `CodecInitializer` to encrypt the stored Token in database.
In this case, when someone accesses the database, he still can't access users accounts until he get access to the `codecKey.txt` file from server.

The `hashedRandomString` has to be encrypted as well IMO and never add it in cookie or expose it as plain text. (it's similar to the password itself).

webui/portal/src/main/java/org/exoplatform/portal/webui/login/UIForgetPassword.java
```diff
@@ -126,8 +127,7 @@ public void execute(Event<UIForgetPassword> event) throws Exception {
 
             // Create token
             RemindPasswordTokenService tokenService = uiForm.getApplicationComponent(RemindPasswordTokenService.class);
-            Credentials credentials = new Credentials(user.getUserName(), "");
-            tokenId = tokenService.createToken(credentials,tokenService.FORGOT_PASSWORD_TOKEN);
+            tokenId = tokenService.createToken(user.getUserName(), CookieTokenService.FORGOT_PASSWORD_TOKEN);
```
boubaker: This class isn't used anymore, shouldn't we delete it ?

component/web/security/pom.xml
```diff
@@ -61,6 +61,10 @@
       <groupId>org.gatein.captcha</groupId>
       <artifactId>simplecaptcha</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk15on</artifactId>
+    </dependency>
```
boubaker: Can you please move this into module `exo.portal.component.api` since it's a generic dependency used by multiple modules.

```suggestion
```

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
rdenarie: ExtentedAttributeManager extends AttributeManagerImpl, which need IdentitySessionImpl in constructor. So ExtendedAttributeManager have to use a constructor with IdentitySessionImpl, to call super(session).

For this reason, I think that it is not possible to change it to a service. 
As you advise, I change in PicketLinkIDMService by reusing the same object, instead of instanciate one each time.

component/identity/src/main/java/org/exoplatform/services/organization/idm/PicketLinkIDMServiceImpl.java
rdenarie: Done.

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
rdenarie: Done.

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
rdenarie: Done

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
rdenarie: Done

component/web/security/src/main/java/org/exoplatform/web/login/RememberMeLoginModule.java
rdenarie: Done

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
rdenarie: Done

webui/portal/src/main/java/org/exoplatform/portal/webui/login/UIForgetPassword.java
rdenarie: Do you mean in import ? 
If yes, I remove it.

component/web/security/pom.xml
rdenarie: Done

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
rdenarie: In picketlink, the encoder is an option of the realm, not of the store.

I understand the concern, and share, but PL seems not allowing it

webui/portal/src/main/java/org/exoplatform/portal/webui/login/UIForgetPassword.java
boubaker: In fact, the whole WebUI class should be removed since it's not used in a WebUI Portlet, isn't it ?

component/identity/src/main/java/org/gatein/portal/idm/impl/store/attribute/ExtendedAttributeManager.java
boubaker: Thanks

webui/portal/src/main/java/org/exoplatform/portal/webui/login/UIForgetPassword.java
rdenarie: I removed the complete 'login' folder

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
boubaker: > In picketlink, the encoder is an option of the realm, not of the store.

ok I see. In PLIDM Ldap Store source code, it seems not used. For my information, did you tested on LDAP conf ?

boubaker (dismissed):

component/identity/src/test/resources/conf/picketlink/exo.portal.component.identity-picketlink-idm-config.xml
rdenarie: I didnt test as LDAP connexion do not hash password

rdenarie: I just test, and the login is ok

rdenarie:
@boubaker 
Can you rewiew one more time ? 
I add a modification to remove unused import and to fix sonar feedback. 
Thanks

boubaker (approved):

PR #648, opened 2023-08-24
feat: Implement "Site management feature" - Meeds-io/MIPs#68

### LINKED ISSUE #68: Site management
## Rationale

No functional requirement for meeds.

In general terms, this MIP allows sites to be managed via an administration function and new sites to be created in a module outside the meeds package.

**Note**: avoid describing the solution here.

### Security 

The new functionalities will be accessible by administrators.

### Extensibility 

We will add the possibility to inject new actions implemented from other addons into Space administration portlet using the extension registry mechanism. 

### Feature Flags

| Property Name  | Default Value | Target Audience | Functional Behaviour |
| ------------- | ------------- | ------------- | ------------- |
|exo.feature.newSiteManagement.enabled  |  true (for eXo), false (for Meeds) |users   |enable/disable the new site management entry in administration menu   |

This feature flag will be disabled by default for Meeds, so the new site management entry will not be displayed for Meeds package.

### Services & processing

- We will implement EditPortalLayout action listener for UIWorkingWorkspace in order to let editing portal site layout by interface.
- We will refactor findSites method of PortalImpl service in order to ignore empty sites.

### Data and persistence

- We will change the character set of description column of PORTAL_SITES table to utf8mb4 in order to support emojis insertion.

azayati:


boubaker (approved):

boubaker:
Just a hint for WebUI extensibility, you can override any component predefined in Meeds using a definition of new factories like done here:
https://github.com/exoplatform/commons/blob/e0afcbbaad150cc9fc20871959b0013c367afc97/commons-webui-component/src/main/resources/META-INF/services/org.exoplatform.portal.application.PortalApplicationFactory
@azayati

PR #653, opened 2023-08-30
feat: Allow to store Registration Settings from UI - MEED-1645 - Meeds-io/MIPs#63

### LINKED ISSUE #63: Hub Access
## Rationale

It is possible to configure a Meeds server as _Opened_ so anyone can sign up. Alternatively, it is also possible to set a Meeds server as _Restricted_ so only known users can sign in. Besides, it's also possible to configure how self registered users will be considered (members or external).

Currently, these settings can only be set by updating platform properties on the server host. As a hub admin, you can't switch these settings via UI. While at meeds, we aim to give capabilities for any Hub Host to manage and administrate their hub on their own as a self service software.

## 1. Functional Requirements

From the administration menu, General/Access settings, 
admin display access settings to precise how users will be able to login into the platform
1. First option:  
* Platform is set as _Open_ by default: anyone can self-register and anyone is considered a full "member" (/platform/user) even when invited from a space.
* -> Ability for the admin to authorize external user when opened. That means that self-registrated users and ones invited by space host will be considered external
* Email signup/signin options is proposed by default and cannot be disabled by UI. Indeed, the authentication topic will be elaborated in another MIP, including metamask, oAuth 

image

2. Second option: 
* Platform is set as _Restricted_  by default: only pre-registered user can access the platform and any user invited from a space is considered external (/platform/external)
* Only email sign in is possible (no sign up)

image

In addition to this, admin will find an easy access to create users from that page so the path to create users is easy and quickly understandable.

And finally, to ease the onboarding of any user accessing the platform the first time, the admin would get an option to set spaces as default for platform/users or platform/externals

> Precision: 
- when switching from a type to another, no automatic update of users. This only has an impact on login access to hub. 
- external access permissions from spaces management is removed as it is here provided in this administration page
- Same for default spaces: no upgrade will be done to previous registered users. This only aims new users
- opened will be the default choice for meeds package. If needed, this choice can be set differently with a property

### Extensibility 

No extensible area nor service is needed for now.

### Configurability 

| Parameter | Default value | Description |
| ------ | ------ | ------ |
| `meeds.settings.access.type.default` | `open` | Possible values : `open` or `restricted`. This will allow to choose the default setting of package. For Meed the access type will be put by default to `open`  | 
| `meeds.settings.access.externalUsers` | `false` | Possible values : `false` or `true`. This will allow to choose the default setting of package. For Meed the registered users will be put by default to `internal`  | 

### Upgradability

No upgrade is needed.

### Existing Features

The configuration, enablement of `Oauth` provider for login shouldn't be impacted by this change. It should remain exactly with the same property settings to be enabled.

### Feature Flags

No Feature Flag is needed.

### Access 

The same Portlet `General Settings` will be used an refactored to allow the display of three Tabs.
A REST endpoint for Hub Access configuration have to be made in order to allow read & write the Hub access settings. This REST endpoint has to be accessible for administrators only (`/platform/administrators`)

### Data and persistence

Since the settings are stored globally for the whole platform. One single entity instance will be needed, thus no need to add a dedicated Storage & DAO layers. The `SettingService` should be used instead to store the Hub Access settings.

boubaker:
This change will introduce a new Service for Registration and Platform Security Settings. This service will determine how a User can sign up into the platform and what is its default permissions.

### DIFF
--- a/component/portal/src/main/java/io/meeds/portal/security/service/SecuritySettingService.java
+++ b/component/portal/src/main/java/io/meeds/portal/security/service/SecuritySettingService.java
@@ -53,21 +53,12 @@ public class SecuritySettingService {
   protected static final String               EXTRA_GROUPS_SEPARATOR             = ",";
 
   protected static final UserRegistrationType DEFAULT_REGISTRATION_TYPE          =
-                                                                        UserRegistrationType.valueOf(System.getProperty("meeds.security.registration.type",
-                                                                                                                        OPEN.name()));
+                                                                        UserRegistrationType.valueOf(System.getProperty("meeds.settings.access.type.default",
+                                                                                                                        OPEN.name()).toUpperCase());
 
   protected static final boolean              DEFAULT_REGISTRATION_EXTERNAL_USER =
-                                                                                 Boolean.parseBoolean(System.getProperty("meeds.security.registration.externalUser",
-                                                                                                                         "false"));
-
-  protected static final String[]             DEFAULT_REGISTRATION_EXTRA_GROUPS  =
-                                                                                Arrays.stream(System.getProperty("meeds.security.registration.extraGroups",
-                                                                                                                 "")
-                                                                                                    .trim()
-                                                                                                    .split(EXTRA_GROUPS_SEPARATOR))
-                                                                                      .filter(StringUtils::isNotBlank)
-                                                                                      .distinct()
-                                                                                      .toArray(String[]::new);
+                                                                                 Boolean.parseBoolean(System.getProperty("meeds.settings.access.externalUsers",
+                                                                                                                         "false").toLowerCase());
 
   private RegistrationSetting                 registrationSetting;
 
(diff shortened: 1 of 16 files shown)


component/portal/src/main/java/io/meeds/portal/security/service/SecuritySettingService.java
```diff
@@ -0,0 +1,174 @@
...
+  protected static final String               REGISTRATION_TYPE_PARAM            = "REGISTRATION_TYPE";
+
+  protected static final String               REGISTRATION_EXTERNAL_USER_PARAM   = "REGISTRATION_EXTERNAL_USER";
+
+  protected static final String               REGISTRATION_EXTRA_GROUPS_PARAM    = "REGISTRATION_EXTRA_GROUPS";
+
+  protected static final String               EXTRA_GROUPS_SEPARATOR             = ",";
+
+  protected static final UserRegistrationType DEFAULT_REGISTRATION_TYPE          =
+                                                                        UserRegistrationType.valueOf(System.getProperty("meeds.security.registration.type",
```
rdenarie: I use the property meeds.security.registration.type to set by default to restricted.
meeds.security.registration.type=restricted

It generate an error at startup : 

> Caused by: java.lang.IllegalArgumentException: No enum constant io.meeds.portal.security.constant.UserRegistrationType.restricted
> 	at java.base/java.lang.Enum.valueOf(Enum.java:273)
> 	at io.meeds.portal.security.constant.UserRegistrationType.valueOf(UserRegistrationType.java:22)
> 	at io.meeds.portal.security.service.SecuritySettingService.<clinit>(SecuritySettingService.java:56)
> 	... 82 common frames omitted
> 

I had to change to 
meeds.security.registration.type=RESTRICTED

As all our properties are not capitalized, could you modify to accept 'restricted' ?

rdenarie (changes requested):

component/portal/src/main/java/io/meeds/portal/security/service/SecuritySettingService.java
boubaker: Ah yes, it must be case insensitive. Done.
BTW, the property names has been changed to fit the ones specified in MIP. Thanks for your help.

rdenarie (approved):

PR #656, opened 2023-09-04
feat: Implement getSites rest endpoint and services in order to retrieve sites with navigations - EXO-65477 - Meeds-io/MIPs#72

IlhemEssaadi:


azayati (approved):
LGTM

PR #658, opened 2023-09-06
feat: Implement getSites rest endpoint and services in order to retrieve sites with navigations - EXO-65477 - Meeds-io/MIPs#72 

azayati:


component/portal/src/main/java/org/exoplatform/portal/mop/dao/SiteDAOImpl.java
```diff
@@ -135,4 +160,67 @@ private List<String> getSiteNames(SiteType siteType, int offset, int limit) {
...
+  private String getQueryFilterName(List<String> suffixes) {
+    return suffixes.isEmpty() ? QUERY_FILTER_FIND_PREFIX : QUERY_FILTER_FIND_PREFIX + "By" + StringUtils.join(suffixes, "And");
+  }
+
+  private <T> void addQueryFilterParameters(SiteFilter filter, TypedQuery<T> query) {
+    if (filter.getSiteType() != null) {
+      query.setParameter(SITE_TYPE, filter.getSiteType());
+    }
+    if (!filter.isAllSites()) {
+      query.setParameter(DISPLAYED, filter.getDisplayed());
```
GouadriaHanen, line 207: In the case of the Boolean type we should check the null case here

component/portal/src/main/java/org/exoplatform/portal/mop/dao/SiteDAOImpl.java
```diff
@@ -135,4 +160,67 @@ private List<String> getSiteNames(SiteType siteType, int offset, int limit) {
...
+
+  private void buildPredicates(SiteFilter filter, List<String> suffixes, List<String> predicates) {
+    if (filter.getSiteType() != null) {
+      suffixes.add("Type");
+      predicates.add("s.siteType = :siteType");
+    }
+    if (!filter.isAllSites()) {
+      suffixes.add("Displayed");
+      predicates.add("s.displayed = :displayed");
+    }
```
GouadriaHanen, line 191: In the case of the Boolean type we should check the null case here

GouadriaHanen (changes requested):

GouadriaHanen (approved):

PR #667, opened 2023-09-21
feat: Redirect after login or access to the uri "/portal" on the default site as the first allowed site if no user homepage defined - EXO-66100 - Meeds-io/MIPs#72

GouadriaHanen:

PR #668, opened 2023-09-25
feat: Redirect after login or access to the uri "/portal" on the default site as the first allowed site if no user homepage defined - EXO-66100 - Meeds-io/MIPs#72 (#667)

GouadriaHanen:


azayati (approved):
LGTM

PR #678, opened 2023-10-11
fix: update indonesian language identifier - EXO-66238 - Meeds-io/meeds#1163

Jihed525:
Before this fix, the indoneasian language is not correctly configured. Since JDK 17, indonesian is 'id', and we use 'in'.
This commit modify the indonesian short identifier to respect jdk17

(cherry picked from commit 4eee46f30fcd0b4af520c0352ee7b04585c7ebc1)


rdenarie (dismissed):
Rework commit message
1) there the customer name is the commit message, we must remove it !
2) Title should describe what the commit fix not the symptome we fix.
3) Commit message do not respect the commit message guideline : https://github.com/Meeds-io/gatein-portal/blob/develop/.github/PULL_REQUEST_TEMPLATE.md?plain=1
It explains the symptome once again, not what is the problem not how it is fix by the commit.
It should be something like : 
Before this fix, the indoneasian language is not correctly configured. Since JDK 17, indonesian is 'id', and we use 'in'.
This commit modify the indonesian short identifier to respect jdk17

or something like that

Jihed525:
I have considered and updated the commit as requested, is it ok @rdenarie ?

rdenarie:
Can you keep the titlz shorter ? No need to speak about jdk 17 in the title as you mention it in the comment

ahamdi (approved):

PR #681, opened 2023-10-12
fix: Fix startup server exception for creating PORTAL_SITES indexes - EXO-66763 - Meeds-io/meeds#1175

GouadriaHanen:
During server startup, two exceptions were raised, one for the "DESCRIPTION" column used in the key specification without key length, since the column type is LONGTEXT, which is not acceptable in MySQL, and an exception for the already existing index key, since the modification set started to run and failed, some indexes were created such as "IDX_PORTAL_SITES_TYPE_01".
After this modification, a precondition was added for indexes already created, and the LONGTEXT index type was removed.


azayati (approved):
LGTM

PR #733, opened 2023-11-14
fix: Fix site properties changing after edit portal layout -  EXO-67453 - Meeds-io/MIPs#88

GouadriaHanen:
Before this fix, after updating the site layout its properties are set to the default value (displayed -> true, displayOrder -> 0, bannerFileId = 0). After this change, we ensure to keep the current values of these properties.


azayati (approved):
LGTM

PR #749, opened 2023-11-24
fix: Add hidden class when portlet is empty-MEED-2977-Meeds-io/meeds#1313

SaraBoutej:
Prior to this change, in all space pages (except the space home page) an extra margin in the top is displayed. This margin is due to the padding-bottom added to all portlet in the platform.
This PR adds the CSS class hidden to the portlet when there is no content to display.

### DIFF
--- a/web/portal/src/main/webapp/groovy/portal/webui/application/UIPortlet.gtmpl
+++ b/web/portal/src/main/webapp/groovy/portal/webui/application/UIPortlet.gtmpl
@@ -144,7 +144,6 @@
 							</div>		
 						</div>
 					</div>
-					
 					<%//Begin Bottom Decorator %>
 					<div class="bottomDecorator clearfix">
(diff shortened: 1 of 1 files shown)


web/portal/src/main/webapp/groovy/portal/webui/application/UIPortlet.gtmpl
```diff
@@ -166,13 +165,17 @@
 			{
 				cssStyle += "height: "+ windowHeight +";";
 			}
+			String hiddenClass = "";
+			if(org.apache.commons.lang.StringUtils.isBlank(portletContent.toString())) {
```
boubaker:
```suggestion
			if(portletContent == null || org.apache.commons.lang.StringUtils.isBlank(portletContent.toString())) {

```

SaraBoutej: done

boubaker (approved):

PR #754, opened 2023-11-27
fix: Fix displaying spaces layout - EXO-67454 - Meeds-io/meeds#1334

GouadriaHanen:
prior to this change, when opening in two tabs a space and administration site, the space page acted like it was a standalone site since the previous site was administrator which is a standalone
after this change, the layout of the space takes the layout of the meta-site after retrieving the dynamicPortalconfig  with the meta-site layout if it is an aggregated site


azayati:
@ahamdi I will wait the commits on the related meeds issue

azayati:
to be closed @GouadriaHanen

PR #763, opened 2023-12-01
feat: Allow to define a Reusable Skin Module - MEED-2994 - Meeds-io/MIPs#108 (#753)

rdenarie:
This change will allow to reuse the same skin module (Client/Browser Path) in different applications. In fact, for a same framework which is used by different applications, we can only define a Portal Skin CSS that will be loaded all time in all pages even when needing it applications aren't present in the page. With this **modularity** feature, we will be able to define a reusable CSS Module that will be injected to page only when a portlet, which needs it, is present to the page.

To define a module, we will be able to use `<filtered>true</filtered>` in a regular portal skin definition to not import it systematically. And then we will be able to reference it in a portlet (multi references) using `<additional-module>NAME<additional-module>*`.

### DIFF
--- a/component/scripting/src/test/resources/UIPortalApplication.gtmpl
+++ b/component/scripting/src/test/resources/UIPortalApplication.gtmpl
@@ -37,6 +37,9 @@
     <%}%>
     <%for(portletSkin in portletSkins) {
     def url = portletSkin.createURL();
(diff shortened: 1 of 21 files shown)


azayati (approved):
LGTM

PR #766, opened 2023-12-05
fix: update database changelog and add a postgresql script - EXO-67619 - Meeds-io/meeds#1373 

ahamdi:
Update changelog to make updates run on the postgresql DB server and avoid blocking following changelogs execution


azayati (approved):
LGTM

PR #772, opened 2023-12-06
feat: Add the possibitily to not authorize the login against the external store - EXO-68030 - meeds-io/meeds#1417

rdenarie:
In some case, when the platform is configured with an external user store AND SSO like OIDC, the IDP have security rules for the login like MFA. But, as the user is present in the external store, he can logs with the eXo login form, bypassing security rules This commit add a property to refused the connection for a user in the external store by the exo login form. He have to use the IDP login form

### DIFF
--- a/component/identity/src/main/java/org/exoplatform/services/organization/idm/externalstore/PicketLinkIDMExternalStoreService.java
+++ b/component/identity/src/main/java/org/exoplatform/services/organization/idm/externalstore/PicketLinkIDMExternalStoreService.java
@@ -40,7 +40,7 @@ public class PicketLinkIDMExternalStoreService implements IDMExternalStoreServic
   public static final Log                                        LOG                           =
                                                                      ExoLogger.getLogger(PicketLinkIDMExternalStoreService.class);
 
-  public static final  IDMEntityType<org.picketlink.idm.api.User> IDMUSER                       =
+  public static final IDMEntityType<org.picketlink.idm.api.User> IDMUSER                       =
                                                                          new IDMEntityType<>("Picket Link IDM User",
                                                                                              org.picketlink.idm.api.User.class);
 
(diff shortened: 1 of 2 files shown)


component/identity/src/main/java/org/exoplatform/services/organization/idm/externalstore/PicketLinkIDMExternalStoreService.java
```diff
@@ -40,7 +40,7 @@ public class PicketLinkIDMExternalStoreService implements IDMExternalStoreServic
   public static final Log                                        LOG                           =
                                                                      ExoLogger.getLogger(PicketLinkIDMExternalStoreService.class);
 
-  public static final IDMEntityType<org.picketlink.idm.api.User> IDMUSER                       =
+  public static final  IDMEntityType<org.picketlink.idm.api.User> IDMUSER                       =
```
ahamdi: Could you format those added lines ?

web/portal/src/main/webapp/WEB-INF/conf/organization/idm-externalstore-configuration.xml
```diff
@@ -38,6 +38,11 @@
         <description>Cron expression used to schedule the job that will import periodically data from external store</description>
         <value>${exo.idm.externalStore.import.cronExpression:0 59 23 ? * *}</value>
       </value-param>
+			<value-param>
```
ahamdi, line 41: As I understand, when enabling MFA, we need to set manually this property too ? couldn't it be activated automatically ?

rdenarie: No.
When I speak about specific security rules like MFA, it is rules managed BY the IDP.

The idea is only to not authorized authentication with exo login form for user in external store (because we want to force them to use openid for example)

component/identity/src/main/java/org/exoplatform/services/organization/idm/externalstore/PicketLinkIDMExternalStoreService.java
rdenarie: done

rdenarie:
cc @ahamdi

ahamdi (approved):

PR #784, opened 2023-12-12
fix: Too many redirect when login - EXO-68055 - Meeds-io/meeds#1423

GouadriaHanen:
Before this change, when login the redirection path is computed from the first accessible site event if its root node is not accessible (restricted navigation) which leads to blank page with too many redirect.
After this commit, after login the redirection path will be computed from the accessible site and navigation nodes with access permission in order to fix the redirection for restricted navigation.


azayati (approved):
LGTM

PR #791, opened 2023-12-19
fix: fix redirecting for global site navigation - EXO-68381 - Meeds-io/meeds#1426

GouadriaHanen:
before this change, when redirecting to a global navigation NPE is thrown since the current site doesn't contain the target navigation
After this change, when the target navigation is not found the redirection is for the requested path


azayati (approved):
LGTM

PR #802, opened 2023-12-29
fix: Fix filtering application after changing permissions - EXO-68546 - Meeds-io/meeds#1505

GouadriaHanen:
Prior to this change, when updating an application permission , its permission is updated only in the UI not in the database. After this change, the permission will be saved in database and the application list will be displayed according to the permissions when selecting a category.


azayati (approved):
LGTM

Jihed525 (approved):

PR #804, opened 2024-01-02
fix: last login time property is not updated in gatein User - EXO-68531 - Meeds-io/meeds#1493

hakermi:
prior to this change, The `UpdateLoginTimeListener` doesn't correctly update the last login property of the user due to double `RequestLifeCycle.begin` that invoked in this listener. 
This PR fixes the listener by removing the `RequestLifeCycle.begin` of the listener and keeps the one invoked by the `@ContainerTransactional` annotation to avoid the start of two transaction lifecycles.


azayati (approved):
LGTM

PR #812, opened 2024-01-15
fix: Containers only acccept pixel or % numeric value for width and height - EXO-66580 - meeds-io/meeds#1527

mkrout:
Prior to this fix, Containers only acccept pixel or % numeric value for width and height, but in sone cases we need to have dynamic container's width as auto or setting a calc expression.
This commit changes the width and height fields validator to allow having more css rules possibilities.

### DIFF
--- a/web/portal/src/main/resources/locale/portal/webui_en.properties
+++ b/web/portal/src/main/resources/locale/portal/webui_en.properties
@@ -254,15 +254,15 @@ UIContainerForm.tab.label.ContainerSetting=Container Setting
 UIContainerForm.tab.label.UIContainerPermission=Permissions
 UIContainerForm.tab.label.Template=Template
 UIContainerForm.tab.label.Icon=#{word.icon}
-UIContainerForm.msg.InvalidWidthHeight=You must enter a pixel or a percentage value in field "{0}".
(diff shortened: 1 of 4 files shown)


sofyenne (approved):

Jihed525 (approved):

web/portal/src/main/resources/locale/portal/webui_en.properties
```diff
@@ -254,15 +254,15 @@ UIContainerForm.tab.label.ContainerSetting=Container Setting
 UIContainerForm.tab.label.UIContainerPermission=Permissions
 UIContainerForm.tab.label.Template=Template
 UIContainerForm.tab.label.Icon=#{word.icon}
-UIContainerForm.msg.InvalidWidthHeight=You must enter a pixel or a percentage value in field "{0}".
+UIContainerForm.msg.InvalidWidthHeight=You must enter a valid css rule in field "{0}".
```
boubaker, line 257: Can you please change the labels on Crowdin instead of PR. (this change will be erased by Crowdin Sync Job next execution if not updated on Crowdin as well @srenault-meeds @Julien-Dubois-eXo  )

PR #815, opened 2024-01-19
fix : Unable to create a user with a username starting with digit - EXO-69012

rdenarie:
Before this fix, the platform refuse to create a user with a username starting with a digit. As there is no technical limitation, this commit change the check about the first char by testing lowerOrDigit instead of just lower


azayati (approved):
LGTM

PR #816, opened 2024-01-22
fix: avoid reusing recycled HTTP requests - Meeds-io/meeds#1565 - EXO-69030

ahamdi:
The issue is related to an old code that tries to get the remote user from the HTTP request, this worked for old Tomcat versions. But since the upgrade to Tomcat 10 and since there is no Security manager configured, the usage of the request object is not permitted after the timeout.
The fix will use COnversationState to get the current user identity

### DIFF
--- a/webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
+++ b/webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
@@ -299,7 +299,9 @@ public class PortalRequestContext extends WebuiRequestContext {
         String remoteUser = "";
         if (userPortalConfig == null) {
             ConversationState conversationState = ConversationState.getCurrent();
-            if(conversationState != null) {
+            if (conversationState != null
+              && conversationState.getIdentity() != null
+              && !IdentityConstants.ANONIM.equals(conversationState.getIdentity().getUserId())) {
               remoteUser = conversationState.getIdentity().getUserId();
             }
             SiteType siteType = getSiteType();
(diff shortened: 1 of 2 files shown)


webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
```diff
@@ -295,8 +296,12 @@ public UserPortal getUserPortal() {
     }
 
     public UserPortalConfig getUserPortalConfig() {
+        String remoteUser = "";
         if (userPortalConfig == null) {
-            String remoteUser = getRemoteUser();
+            ConversationState conversationState = ConversationState.getCurrent();
+            if(conversationState != null) {
```
boubaker: Should check if user is anonymous IMO (even if it doesn't happen now, we may face an improvement that will change to have an "ANONIM" user when not connected, thus just in case...)

```suggestion
            if (conversationState != null
              && conversationState.getIdentity() != null
              && !IdentityConstants.ANONIM.equals(conversationState.getIdentity().getUserId())) {
```

webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
```diff
@@ -295,8 +296,12 @@ public UserPortal getUserPortal() {
     }
 
     public UserPortalConfig getUserPortalConfig() {
+        String remoteUser = "";
```
boubaker: Should be null if not connected IMO

```suggestion
        String remoteUser = null;
```

boubaker (commented):
Thanks, small remarks

webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
ahamdi: Ok

webui/portal/src/main/java/org/exoplatform/portal/application/PortalRequestContext.java
ahamdi: Yes that makes sense

boubaker (dismissed):

ahamdi:
@boubaker Could you re-approve ? thanks

boubaker (approved):

PR #828, opened 2024-02-06
fix: Update GroupPortalConfigListener in order to create group sites only for those respecting group name pattern - EXO-69296 - Meeds-io/meeds#1608

azayati:

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/config/GroupPortalConfigListener.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/config/GroupPortalConfigListener.java
@@ -1,20 +1,20 @@
 /**
- * Copyright (C) 2009 eXo Platform SAS.
+ * This file is part of the Meeds project (https://meeds.io/).
  *
- * This is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


component/portal/src/main/java/org/exoplatform/portal/config/GroupPortalConfigListener.java
```diff
@@ -37,118 +39,100 @@
...
-    public void postSave(Group group, boolean isNew) throws Exception {
-        if (!isNew) {
-            return;
-        }
-
-        RequestLifeCycle.begin(PortalContainer.getInstance());
-        try {
-            String groupId = group.getId();
-
-            // Bug in hibernate org service implementation
```
rdenarie, line 87: Are you sure this is no more needed ?

azayati: I have tried without, it works fine, I can keep it to be more secure

rdenarie (dismissed):

rdenarie (approved):

PR #834, opened 2024-02-13
fix: Fix User Identity Update On First Login - MEED-3337 - Meeds-io/meeds#1620 (#833)

rdenarie:
Prior to this change, the Current User Identity isn't refreshed right away after applying default memberships. This change will ensure to refresh identity in Identity Registry just after adding memberships and to update the last login time in the same process to ensure not adding the user twice in the same default groups.


azayati (approved):
LGTM

rdenarie:
Not reproduced on 6.5.x => no backport

PR #838, opened 2024-02-21
feat: allow to parse the DN and get the username when LDAP property is of type user - EXO-69802 - Meeds-io/MIPs#112  

ahamdi:
When synchronizing users from LDAP, all attribute values are considered as text fields.
When the attribute is of type user, the returned value is the DN of the user, which is not a useful information on the platform.
The improvement will check the type of the property while importing the attribute and if it is of type user, it will extract the user ID from it respecting the configuration of LDAP mappings.

### DIFF
--- a/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
+++ b/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
@@ -3,11 +3,13 @@ package org.picketlink.idm.impl.store.ldap;
 import org.apache.commons.lang3.StringUtils;
 import org.picketlink.idm.common.exception.IdentityException;
 import org.picketlink.idm.impl.NotYetImplementedException;
+import org.picketlink.idm.impl.api.SimpleAttribute;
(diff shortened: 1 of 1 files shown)


component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
```diff
@@ -565,4 +562,120 @@ private LDAPIdentityObjectImpl getSafeLDAPIO(IdentityStoreInvocationContext ctx,
...
+
+            // check if the value is the DN of another identity type
+            IdentityObject identityObject = findIdentityObject(ctx, value);
+            // If it is an identity object, let's add its ID
+            if(identityObject != null) {
+              identityObjectAttribute.addValue(identityObject.getName());
+            } else { // Otherwise, we add the String value as is
+              identityObjectAttribute.addValue(value);
+            }
+          }
```
ahamdi, line 634: This is the added code

rdenarie: seems ok for me

PR #845, opened 2024-02-28
fix: fix sending email with user local when using user email - EXO-69927 - Meeds-io/meeds#1705

GouadriaHanen:
before this change, when using the user email as userId in an email notification, the email is sent in EN even user's language setting is FR since it could not find a user by name with an email
After this change, when userId contains an @ we search for the user with its email


mkrout (dismissed):

rdenarie (changes requested):
Hello,
In which cases getUserLocale is called with a userId containing an arobase ?

rdenarie:
I check the task, 
As I understand, this is called in agenda, when I invite someone by his email. 

As the search by email is time consuming, and not very performant, I think we should transform the email in a user name in agenda module, BEFORE creating the notification. 

getUserLocale called with a mail instead a username must occurs in only one case : the email is an external, and have no account in the platform

GouadriaHanen:
as proposed by @rdenarie this PR will be closed

PR #850, opened 2024-03-21
fix: Fix breadcrumb portlet redirection in spaces - EXO-70408 - Meeds-io/meeds#1789

GouadriaHanen:
Before this change, when clicking on a breadcrumb item of a space we have a wrong redirection since the URL is computed with the wrong space.
After this change, the URL is computed with the right space node URL and the redirection is done successfully.


azayati (approved):
LGTM

PR #858, opened 2024-04-01
fix: Unsearchable disabled users by their email or name - EXO-69348 - Meeds-io/meeds#1827 .

akhanfir:
Before this change, when set user as a disabled user, select disabled filter and search for usera by his name or his email, user isn't displayed in the result list. To fix this problem, first apply the search by email and then change in the search query the conditions apply in the `WHERE` line, they are based on the fields email and username must be both like the search term but with this change will be based on one of the two fields either being like this term and this change will only be applied if the search done by the keyword and the user status is disabled.
After this change, user is displayed in the result list.


rdenarie (dismissed):
Hello, 
Can you complete the commit message explaining how you fix the problem ?
While Reading the modifications, I'm not able to understand your fix

mkrout (approved):

PR #861, opened 2024-04-02
fix: login with unexisting user lead to a white page - EXO-70765

### LINKED ISSUE #1826: Login with not existing user displays blank page
**Steps to reproduce**

1. Go to Login page
2. Try to login with non existing user
3. See page result

**Current behaviour**

Empty blank page is displayed

**Expected behavior**

Login page should be displayed with error message to inform that login doesn't succeeded 

**(optional) Affected environments**  
Version: 1.6.x-SNAPSHOT (develop)

rdenarie:
Before this fix, when the user logs with a non existing user, a white page is displayed instead of the login page with the classic error message This problem is due to a test when caseInsensitive is true, which set the username to a null value This commit add an additional test to not set the username to null

Resolves meeds-io/meeds#1826

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
@@ -200,7 +200,7 @@ public class LoginHandler extends JspBasedWebHandler {
               username = realUsername;
             }
           }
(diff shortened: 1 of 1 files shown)


azayati (approved):
LGTM

PR #862, opened 2024-04-02
fix: login with unexisting user lead to a white page - EXO-70765

rdenarie:
Before this fix, when the user logs with a non existing user, a white page is displayed instead of the login page with the classic error message This problem is due to a test when caseInsensitive is true, which set the username to a null value This commit add an additional test to not set the username to null

Resolves meeds-io/meeds#1826


azayati (approved):
LGTM

PR #866, opened 2024-04-10
fix: [SAML] Some properties cannot be modified by configuration - EXO-70965

rdenarie:
Before this fix, property gatein.sso.saml.config.file is not taken in account, the value used was always ${exo.conf.dir}/saml2/picketlink-sp.xml This is due to a default value in context.xml, which not use the property This fix modifies this file to correctly use the property

Resolves meeds-io/meeds#1856


azayati (approved):
LGTM

PR #880, opened 2024-05-03
fix: Update to jdk21 - EXO-71474 - Meeds-io/MIPs#91

### LINKED ISSUE #91: feat: Upgrade to JDK 21
## Rationale

JDK 21 is the latest LTS version and we should migrate into it to benifit from its dvantages and to keep possibility of supporting a "supported JDK" for customers contexts.

### Feature Flags

| Property Name  | Default Value | Target Audience | Functional Behaviour |
| ------------- | ------------- | ------------- | ------------- |
|  |   |   |   |

### Access 
   - GUI 
   - API 

### Services & processing
   - Ex: Scheduled Jobs

### Data and persistence
   - Macro data model
   - Persistence layer 
   - Use of caches
## 5. Annexes
   - Impacts 
   - References

rdenarie:
Before this fix, the simpleDateFormat for locale en is MM/dd/yyyy, HH:mm:ss a
In jdk17 and under, the space between 'ss' and 'a' is a simple space In JDK 21, the space was changed to '[NNBSP]a' ([OpenJDK information](https://bugs.openjdk.org/browse/JDK-8324665)) : , so the .contains(" a") returns always false

This commit add the test on '[NNBSP]a' in addition of ' a' in order to make it work on both jdk version

Resolves Meeds-io/MIPs#91

### DIFF
--- a/component/common/pom.xml
+++ b/component/common/pom.xml
@@ -93,6 +93,12 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>@{argLine} ${env.MAVEN_OPTS} --add-opens=java.base/jdk.internal.vm=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED</argLine>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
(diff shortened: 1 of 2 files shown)


component/common/pom.xml
```diff
@@ -93,6 +93,12 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>@{argLine} ${env.MAVEN_OPTS} --add-opens=java.base/jdk.internal.vm=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED</argLine>
```
boubaker, line 99: May be to add here https://github.com/Meeds-io/maven-parent-pom/blob/38f428610ced8e1824eb72d9dd35c5635b771fc8/pom.xml#L768 WDYT ?

boubaker: We may need to add those options here as well (for runtime):
https://github.com/Meeds-io/meeds/blob/88622dd72c45a772a792311350c15c606cb0c09d/packaging/plf-tomcat-resources/src/main/resources/bin/setenv.sh#L293

( for windows here https://github.com/Meeds-io/meeds/blob/88622dd72c45a772a792311350c15c606cb0c09d/packaging/plf-tomcat-resources/src/main/resources/bin/setenv.bat#L199 )

PR #888, opened 2024-05-16
Merge mip 91

rdenarie:

### DIFF
--- a/component/common/pom.xml
+++ b/component/common/pom.xml
@@ -93,6 +93,12 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>@{argLine} ${env.MAVEN_OPTS} --add-opens=java.base/jdk.internal.vm=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED</argLine>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
--- a/component/common/pom.xml
+++ b/component/common/pom.xml
@@ -93,12 +93,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <argLine>@{argLine} ${env.MAVEN_OPTS} --add-opens=java.base/jdk.internal.vm=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED</argLine>
-        </configuration>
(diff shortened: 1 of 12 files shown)


component/common/pom.xml
```diff
@@ -93,6 +93,12 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>@{argLine} ${env.MAVEN_OPTS} --add-opens=java.base/jdk.internal.vm=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED</argLine>
```
boubaker: Should we generalize this for all Unit Test executions ? If yes, I think it would be better to mode this declaration in centralized one [here](https://github.com/Meeds-io/maven-parent-pom/blob/38f428610ced8e1824eb72d9dd35c5635b771fc8/pom.xml#L768)

(And update runtime scripts setenv.sh & setenv.bat )

rdenarie: This configuration is not exactly the same for all modules. 
So, we can let as is so each module is configured with the lesser possible argument, and it can be easily updated if needed, or we can centralize it, and in this case, all modules will have same configuration, but with some elements not needed. In addition, if we need to update it, we need to release the parent pom.

For me, I prefer to let it in each modules

For setenv.sh I'm not sure to understand what you mean

boubaker: `--add-opens=java.base/jdk.internal.vm=ALL-UNNAMED`
If this is needed in one module, this means it will be needed in any module calling this one which needs to access `jdk.internal.vm`. I 'm ok to add special parameters in modules for tests but I prefer to add this kind of parameters everywhere once any module needs to declare it which will avoid exceptions in upper modules depending on Unit Tests of gatein-portal.
For `setenv.sh` and `setenv.bat`, we declare also the `--add-opens`. So if we detected during Unit Tests that we need to add more parameters of type `--add-opens`, it's propably needed as well in runtime, thus adding it is better to have exception after deploying in productions, [here](https://github.com/Meeds-io/meeds/blob/88622dd72c45a772a792311350c15c606cb0c09d/packaging/plf-tomcat-resources/src/main/resources/bin/setenv.sh#L293):
```
JDK_JAVA_OPTIONS="$JDK_JAVA_OPTIONS --add-opens java.base/java.io=ALL-UNNAMED"
JDK_JAVA_OPTIONS="$JDK_JAVA_OPTIONS --add-opens java.base/java.lang=ALL-UNNAMED"
JDK_JAVA_OPTIONS="$JDK_JAVA_OPTIONS --add-opens java.base/java.lang.invoke=ALL-UNNAMED"
JDK_JAVA_OPTIONS="$JDK_JAVA_OPTIONS --add-opens java.base/java.lang.reflect=ALL-UNNAMED"
JDK_JAVA_OPTIONS="$JDK_JAVA_OPTIONS --add-opens=java.base/java.lang.module=ALL-UNNAMED"
```

This is the strategy that we adopted during last upgrades of JDK BTW.

rdenarie: ok, I'll do modifications

Treasureth:
Great job here

boubaker (approved):

PR #899, opened 2024-05-29
fix : fill cache of InMemoryGroupDaoHandler with a Modifiable Map - EXO-72017

rdenarie:
Before this fix, when we call computeIfAbsent in InMemoryGroupDaoHandler, and when the item is not in cache, then the create map is done with Collections.emptyMap(). As this map is a non modifiable map, it is not possible to create a group in a test after that as it throws a UnsupportedExecption

This commit change the map to a classic empty Hashmap, so that it can be updated in the future

Resolves meeds-io/meeds#2068

### DIFF
--- a/component/identity/src/test/java/org/exoplatform/services/organization/mock/InMemoryGroupHandler.java
+++ b/component/identity/src/test/java/org/exoplatform/services/organization/mock/InMemoryGroupHandler.java
@@ -175,7 +175,7 @@ public class InMemoryGroupHandler implements GroupHandler {
 
   @Override
   public ListAccess<Group> findGroupChildren(Group parent, String keyword) {
-    List<Group> childGroups = groupChildsById.computeIfAbsent(parent.getId(), key -> Collections.emptyMap())
+    List<Group> childGroups = groupChildsById.computeIfAbsent(parent.getId(), key -> new HashMap<>())
                                              .values()
                                              .stream()
                                              .filter(group -> StringUtils.contains(group.getLabel(), keyword)
@@ -185,7 +185,7 @@ public class InMemoryGroupHandler implements GroupHandler {
   }
 
   public Collection<Group> findGroups(Group parent) {
-    return groupChildsById.computeIfAbsent(parent.getId(), key -> Collections.emptyMap())
+    return groupChildsById.computeIfAbsent(parent.getId(), key -> new HashMap<>())
                           .values()
                           .stream()
                           .map(ObjectUtils::clone)


component/identity/src/test/java/org/exoplatform/services/organization/mock/InMemoryGroupHandler.java
```diff
@@ -175,7 +175,7 @@ public Group findGroupById(String groupId) {
 
   @Override
   public ListAccess<Group> findGroupChildren(Group parent, String keyword) {
-    List<Group> childGroups = groupChildsById.computeIfAbsent(parent.getId(), key -> Collections.emptyMap())
+    List<Group> childGroups = groupChildsById.computeIfAbsent(parent.getId(), key -> new HashMap<>())
```
boubaker, line 178: a `find...` method should never return a modifiable collection that **gives direct access** to cahce entries

```suggestion
    List<Group> childGroups = groupChildsById.computeIfAbsent(parent.getId(), key -> Collections.emptyMap())
```

rdenarie: Here is my pb : 
In my test, I firstly do a call to 
findGroupByChildren : https://github.com/Meeds-io/gatein-portal/blob/2245c74981a4a302dcb436d6aefded12cde89e87/component/identity/src/test/java/org/exoplatform/services/organization/mock/InMemoryGroupHandler.java#L177-L185
At this moment, the parent have no children, so the computeIfAbsent execute, and add an emptyMap (unmodifiable in groupChildsById.

Later, I try to add a children to the parent by calling saveGroup
So this is executed : 
https://github.com/Meeds-io/gatein-portal/blob/2245c74981a4a302dcb436d6aefded12cde89e87/component/identity/src/test/java/org/exoplatform/services/organization/mock/InMemoryGroupHandler.java#L123-L124
as the group.parentId exists, the computIfAbsent is not executed, and the put is launched on the unmodifiable map, which lead to UnsupporedOperationException

So I'm not sure to what I should do here.

Last but not least : if I call saveGroup, before the find, the list put in groupChildsById is a regular Hashmap, not a unmodifiable one.

What is your proposition on this point ?

boubaker: I see, thanks for the explanation. So you doesn't want to change the result of `find...` bu to change the `Map` stored in Cache.

No problem with that, just a little remark about title:
> When filling the cache in InMemoryGroupDaoHandler, do not return a non modifiable map
May be to change to
> fill cache of InMemoryGroupDaoHandler with a Modifiable Map

rdenarie: Ok, I change the description

boubaker (approved):

PR #955, opened 2024-09-04
feat: Define eXo.env.portal.availableLanguages eXo variable as the current available languages - EXO-65763 - Meeds-io/MIPs#128 

sofyenne:


azayati (approved):
LGTM

PR #956, opened 2024-09-04
Merge meeds-io/MIPs#128

sofyenne:


web/portal/src/main/webapp/groovy/portal/webui/workspace/UIPortalApplication.gtmpl
```diff
@@ -181,6 +200,7 @@
      eXo.env.portal.userName='<%= userName == null ? "" : userName %>';
      eXo.env.portal.language='<%= language %>';
      eXo.env.portal.defaultLanguage='<%= defaultLocaleTag %>';
+     eXo.env.portal.availableLanguages='<%= localesJSON.toString() %>';
```
boubaker, line 203: Since this object is huge and its usage is in a single place, it should be better to place it in the JSP of the portlet(s) where it's needed only without affecting other pages loading and Browser Memory consumption.

boubaker: BTW, it should even better to reuse the endpoint of TranslationService.js to retrieve such an information:

https://github.com/Meeds-io/social/blob/a959703f9f2939af02da0090da3757386a6aaa3f/webapp/portlet/src/main/webapp/vue-apps/component-translation-field/js/TranslationService.js#L20-L27

```javascript
export function getTranslationConfiguration() {
  if (translationConfiguration) {
    return Promise.resolve(JSON.parse(JSON.stringify(translationConfiguration)));
  } else {
    return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/social/translations/configuration`, {
      method: 'GET',
      credentials: 'include',
    }).then((resp) => {
      if (resp?.ok) {
        return resp.json();
      } else {
        throw new Error('Error when getting translation configuration');
      }
    }).then((configuration) => {
      translationConfiguration = configuration;
      return JSON.parse(JSON.stringify(translationConfiguration));
    });
  }
}

```

which returns the default platform language as configured in `Main Settings` page and the supported locales:
image

azayati: ok we will right now use an existing rest service from notes, then in the next MIP enhance that with the TranslationService.js since it needs more code adaptation

PR #968, opened 2024-09-13
fix: Update textColor and greyColorLighten1 branding variables default values - EXO-74000 - Meeds-io/MIPs#129

hakermi:
Update textColor and greyColorLighten1 variables default values


rdenarie (dismissed):
We should not change theses colors !

rdenarie (changes requested):
Does the value are eXo values ? or meeds and exo values ?

hakermi:
> Does the value are eXo values ? or meeds and exo values ?

This affects eXo only because of the deployment order of modules, in meeds this get overridden by Platform-ui,
in eXo it gets overridden by digital-workplace

PR #1018, opened 2024-12-18
fix: Missing double quote around component class - EXO-76079

rdenarie:
This commit add a missing double quote in the template which generate DOM error on page display

Resolves https://github.com/Meeds-io/meeds/issues/2690

### DIFF
--- a/web/portal/src/main/webapp/groovy/portal/webui/container/UITableColumnContainer.gtmpl
+++ b/web/portal/src/main/webapp/groovy/portal/webui/container/UITableColumnContainer.gtmpl
@@ -13,7 +13,7 @@
   String uiComponentClass = uicomponent.getCssClass() == null ? "" : uicomponent.getCssClass();
 %>
 <div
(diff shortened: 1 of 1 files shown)


Jihed525 (approved):

web/portal/src/main/webapp/groovy/portal/webui/container/UITableColumnContainer.gtmpl
```diff
@@ -13,7 +13,7 @@
   String uiComponentClass = uicomponent.getCssClass() == null ? "" : uicomponent.getCssClass();
 %>
 <div
-  class="UIContainer UITableColumnContainer <%=uiComponentClass%>  id="${uicomponent.id}">
+  class="UIContainer UITableColumnContainer <%=uiComponentClass%>" id="${uicomponent.id}">
```
boubaker, line 16: It's not the only `gtmpl` file in fact, I 'll take care of updating the others as well

rdenarie: ok, thanks

PR #1076, opened 2025-03-03
fix: Apply data type change for CUSTOMIZATION field of PORTAL_WINDOWS table only with "mysql" dbms - MEED-8510 - Meeds-io/meeds#3055

azayati:
Prior to this change, the PORTAL_WINDOWS table changeset changing the data type of CUSTOMIZATION field to "LONGBLOB" is not working for postgres dbms which causes some errors for postgres server startup. After this change, we apply this changeset only for mysql dbms case since for postgres dbms the CUSTOMIZATION field data type is already "oid" so no need to change it.

### DIFF
--- a/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
+++ b/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
@@ -403,7 +403,8 @@
       </sql>
   </changeSet>
 
-  <changeSet author="portal" id="1.0.0-42">
+  <changeSet author="portal" id="1.0.0-42" dbms="mysql">
+    <validCheckSum>ANY</validCheckSum>
     <modifyDataType tableName="PORTAL_WINDOWS" columnName="CUSTOMIZATION" newDataType="LONGBLOB" />
   </changeSet>
 


mkrout (approved):

component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml
```diff
@@ -403,7 +403,8 @@
       </sql>
   </changeSet>
 
-  <changeSet author="portal" id="1.0.0-42">
+  <changeSet author="portal" id="1.0.0-42" dbms="mysql">
```
boubaker, line 406: Does this mean that `CUSTOMIZATION` column will not have enough length as much as when using MySQL ?
IMO, we should update the `CUSTOMIZATION` column on PostgreSQL as well ([exemple of solutions](https://stackoverflow.com/a/52297510))

azayati: @boubaker when we return to bytea for pgsql, we have the same errors [here](https://github.com/Meeds-io/meeds/issues/692) which is fixed by this [change](https://github.com/Meeds-io/gatein-portal/commit/2a532a6bba2c510d6d3d218606424875213a2b50)
IMO in our case, the oid type is the adequate pgsql type which corresponds to LONGBLOB

boubaker: > @boubaker when we return to bytea for pgsql, we have the same errors [here](https://github.com/Meeds-io/meeds/issues/692) which is fixed by this [change](https://github.com/Meeds-io/gatein-portal/commit/2a532a6bba2c510d6d3d218606424875213a2b50) IMO in our case, the oid type is the adequate pgsql type which corresponds to LONGBLOB

ok, what's the solution IYO ?

rdenarie: oid is the type for large object in psql (Object Identifier). 
So the oid type already accept large binary values. 
So, we update the type to longblob for mysql, and keep oid for psql.

When using blob in liquibase, hibernate interpret it and use oid type in psql. Before Hibernate update in meeds 1.3.x, the type used was bytea. So, Hibernate force us to keep oid as equivalent of blob in psql.

azayati: > > @boubaker when we return to bytea for pgsql, we have the same errors [here](https://github.com/Meeds-io/meeds/issues/692) which is fixed by this [change](https://github.com/Meeds-io/gatein-portal/commit/2a532a6bba2c510d6d3d218606424875213a2b50) IMO in our case, the oid type is the adequate pgsql type which corresponds to LONGBLOB
> 
> ok, what's the solution IYO ?

IMO except any other constraint, keep oid as data type for the customization field since it is dedicated to large objects storage and interpreted correctly by hibernate

boubaker: > oid is the type for large object in psql (Object Identifier). So the oid type already accept large binary values. So, we update the type to longblob for mysql, and keep oid for psql.
> 
> When using blob in liquibase, hibernate interpret it and use oid type in psql. Before Hibernate update in meeds 1.3.x, the type used was bytea. So, Hibernate force us to keep oid as equivalent of blob in psql.

ok, I see, so both Databases will have the same max length of `4,294,967,295 bytes` due to changes introduced for Postgres using [this liquibase changeset](https://github.com/Meeds-io/gatein-portal/blob/825e4012aaaada7ee2720a833abdee472c38555f/component/portal/src/main/resources/db/changelog/portal-rdbms.db.changelog-1.0.0.xml#L277-L289). So it wasn't the case in 1.5.x, but here in 7.0 we will have an aligned size of data type in both supported databases. It's ok for me.

We may need to add **this conclusion** inside a comment of changed liquibase changeset to provide a better understanding of Columns Data Types with its coherence for a better evolutivity and maintainability. By example:

```suggestion
  <!-- Change 'CUSTOMIZATION' column Length in MySQL only in order to correspond to used Column length in Postgres which uses OID Data Type -->
  <changeSet author="portal" id="1.0.0-42" dbms="mysql">
```

WDYT ?

azayati: yes it is ok for me @boubaker thanks

PR #1089, opened 2025-04-04
fix: just after creation, external user is seen as internal user - MEED-8716

rdenarie:
Before this fix, after external user creation, he is automatically loggued into the platform and have outdated rights : he is still platform/users (but not in database)

During external user creation, we have 3 steps
1. Create user (he is automatically added in /platform/users by listeners)
2. deleteFromInternalUsersGroup (remove him from /platform/users)
3. addToExternalUsersGroup (add him in /platform/externals. This action runs SocialMembershipListenerImpl which add user in space he is invited)

The step 3. fire an asynchronous listener GamificationGenericListener on event joinSpace. This listener read user membership in database and put it in cache.

Currently, when the GamificationGenericListener read membership, the /platform/users group is still not persisted in database, and so, see the user as member of /platform/users. As membership are cached, when the user logs in, memberships are still /platform/users

This commit ensure to restart the transaction after deleteFromInternalUsersGroup and before addToExternalUsersGroup, so that deletion of membership /platform/users is persisted before step.

### DIFF
--- a/component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
@@ -45,6 +45,7 @@ import jakarta.servlet.http.HttpSession;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.exoplatform.container.component.ComponentRequestLifecycle;
 import org.json.JSONObject;
 
 import org.exoplatform.commons.utils.ListAccess;
@@ -402,6 +403,7 @@ public class ExternalRegisterHandler extends JspBasedWebHandler {
     if (!isAdministrator && securitySettingService.isRegistrationExternalUser()) {
       // Avoid incoherence by indicating an admin user As external
       deleteFromInternalUsersGroup(login);
+      restartTransaction();
       addToExternalUsersGroup(login);
     }
     return login;
@@ -641,5 +643,4 @@ public class ExternalRegisterHandler extends JspBasedWebHandler {
     user.setPassword(password);
     return user;
   }
-
 }
--- a/component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
+++ b/component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
@@ -45,7 +45,6 @@ import jakarta.servlet.http.HttpSession;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.exoplatform.container.component.ComponentRequestLifecycle;
 import org.json.JSONObject;
 
 import org.exoplatform.commons.utils.ListAccess;


component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
```diff
@@ -401,8 +402,16 @@ private String createUser(User user) throws Exception {
     boolean isAdministrator = CollectionUtils.isNotEmpty(memberships);
     if (!isAdministrator && securitySettingService.isRegistrationExternalUser()) {
       // Avoid incoherence by indicating an admin user As external
-      deleteFromInternalUsersGroup(login);
-      addToExternalUsersGroup(login);
+      begin(organizationService);
```
boubaker: IMO, it would be better call a transitive transaction to ensure that it englobes a valid transaction for IDM listeners as well which may need a transaction on a different Datasource.

```suggestion
      RequestLifecycle.begin(ExoCotnainerContext.getCurrent...());
```

boubaker: BTW, a transaction should already be started due to this flag: `org.exoplatform.web.register.ExternalRegisterHandler.getRequiresLifeCycle()` which returns `true`

boubaker: If the transaction isn't started, we just need IMO to make a fix in `org.exoplatform.web.WebAppController.service(HttpServletRequest, HttpServletResponse)` rather than this specific Web Handler only. WDYT ?

rdenarie: In fact, I wish to force the organizationService and Picketlink to persist the modifications. 
And I see that, in wrapForAutomaticLogin we have a restartTransaction. 

So I have to recheck this part

boubaker: ok, just to ensure that as result the IDM connection is closed, you check that the IDM Datasource has 0 active connections using jconsole on Datasource MBean.

hint: the restartTransaction can not work as expected due to a potential falsy condition in `org.exoplatform.web.WebAppController.service` (the returned container of `ExoContainerContext.getCurrentContainer()` has to be of type `PortalContainer`, else the IDM transaction isn't started, to check by debug mode if needed)

image

rdenarie:
@boubaker I update the commit and the fix explaination.

I used restartTransaction between deleteFromInternalUsersGroup and addToExternalUsersGroup to force persist modification in DB.

restartTransaction  has been tested in debug mode, and works as expected

component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
```diff
@@ -45,6 +45,7 @@
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.exoplatform.container.component.ComponentRequestLifecycle;
```
boubaker:
```suggestion
```

boubaker (commented):
Thanks, just an unused import

component/web/security/src/main/java/org/exoplatform/web/register/ExternalRegisterHandler.java
rdenarie: done

boubaker (approved):

PR #1149, opened 2025-07-17
Merge Mip 203

### LINKED ISSUE #203: Login Portlets & Pages
## Rationale

Login related pages are not customizable yet and multiple times, it happens that CSS is applied to personnalize labels, to add contents above the login form. 
Indeed, as for example, when the login page suggest an additional login method in addition to the meeds login form (OIDC, SAMLv2, etc.), then the labels are not understandable enough for users.
Hence, they sometimes try to enter their credentials in the Meeds Login Form instead of clicking to the SSO authentication CTA.

In order to avoid adding CSS to these pages and to help end-users accessing the platform the easiest as possible, we thus imagine this MIP to enable admins customize any related login pages:
- login page
- registration page
- forgot password page

Please find below the details.

#### Login portlets

Different portlets need to be created, based on what it is currently provided:
- Login form

Image

- Registration form

Image

- Forgot password form

Image

In addition, the user journey when multiple authentication methods are provided will be reviewed:
- as far as there is one authentication method is provided in addition to the meeds login form
- then the login form will not be provided
- a CTA will be suggested to finally access the login form
This must improve the experience when multiple authentication are listed.

Image

Each of these portlets are listed as available in instances in a new category called "Login". Only platform admins are able to use those ones.

#### Portlets customization

Admins will only have customization options for login portlet in order to review the text, the options to set the authentication options.
Other portlet will only be a portlet so it is editable using layout options.

External providers will be provided depending on the configuration.

Image

#### Other portlets

To enable admins to reuse items listed defaultly in login pages, other portlets are needed:
- Platform Name:
    - with default style of title
    - if needed, the style can be changed using layout options
- Platform Logo:
    - different than topbar logo as we only need the platform logo and not the name
    - if needed, the style can be changed using layout options (margin, radius, background, etc.)
- Sidebar login:
   - App displayed at the left of the desktop view of login page
   - Settings for this app will be provided in a drawer thanks to a settings access for admin when hovering it

#### Pages management

In order to let admins edit the layout of each page, those ones will be listed in the global site with the option to edit it if needed.

Each page is defaultly organized the same in a grid view
- Page design specifications:
   - full window
   - no margins
- Design to be adapted after some tests using portlets

#### Review of login page customization page

Since login page can be customized using layout management and global pages, then the login page customization page is suggesting the edition of each page.

Image

#### Gamification
NA

#### Notifications
NA

#### Analytics
NA

#### Unified Search
NA

### Security 

Login, Register and Forgot password portlet will be accessible by anonymous users.

### Existing Features

Currently, the url /portal/login is managed by LoginHandler, which allow to display the login form, and to capture the authentication flow. 

We will add a new handler : PageAccessHandler. This handler will allow to access to a page of global site directly, if permissions allows.

With this handler, when accessing to url /portal/login, will display page login in portal global, if permissions are suitable.
This handler can be used for all pages of global portal, for all levels in navigation.

This login handler will be kept. If the request is not a POST containing parameters needed for the authentication, it do nothing, and let the other handler process the request.
Else, the LoginHandler still process the authentication, we will not change the login handler which will continue to perform the authencation, or delegate it to SSO configurations.

The pageAccessHandler will be used for other page : Register and onboarding

### Feature Flags

No feature flags. Portlets will be created in step 1, without adding it to pages, then we conserve current authentication process. 
When portlets are ready, we will process to step 2 : creating the new login page.

### Security
The new pages created in global portal must be accessible to all users.

### Access 
   - GUI 
   - API 

### Services & processing
   - Ex: Scheduled Jobs

### Data and persistence

New portlets will be created and will handle settings in Portlets preferences

## 5. Annexes
   - Impacts 
   - References

rdenarie:
This commit add the portlet Login Form


component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
```diff
@@ -218,6 +218,19 @@ public String getContextPath() {
         } catch (Exception e) {
           LOG.debug("User {} authentication failed.", username, e);
           status = LoginStatus.FAILED;
+          String referer = request.getHeader("Referer");
+          if (referer!=null && !referer.contains("/portal/login?")) {
+            // If the request comes from a different page, we redirect to the login page
+            // with the error message
+            try {
+              String redirectUrl = referer.contains("?") ? "&" : "?";
+              redirectUrl = referer+redirectUrl+"error=" + status.getErrorCode();
```
boubaker:
```suggestion
              redirectUrl = "%s%serror=%s".formatted(referer, redirectUrl, status.getErrorCode());
```

component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
```diff
@@ -218,6 +218,19 @@ public String getContextPath() {
         } catch (Exception e) {
           LOG.debug("User {} authentication failed.", username, e);
           status = LoginStatus.FAILED;
+          String referer = request.getHeader("Referer");
+          if (referer!=null && !referer.contains("/portal/login?")) {
```
boubaker: Code format


```suggestion
          if (referer != null && !referer.contains("/portal/login?")) {
```

component/web/security/src/main/java/org/exoplatform/web/login/onboarding/OnboardingHandler.java
```diff
@@ -305,4 +285,14 @@ private void writeImage(HttpServletResponse response, BufferedImage bi) {
       log.error(e.getMessage(), e);
     }
   }
+
+  private static void returnSubmissionError(String message, HttpServletResponse response) throws
+                                                                                                        IOException {
```
boubaker: Code format

webui/src/main/java/org/exoplatform/portal/application/PageRequestHandler.java
```diff
@@ -0,0 +1,344 @@
...
+import org.exoplatform.web.application.RequestFailure;
+import org.exoplatform.web.controller.QualifiedName;
+import org.exoplatform.webui.core.UIApplication;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.ServiceLoader;
+
+@SuppressWarnings("rawtypes")
+public class PageRequestHandler extends WebRequestHandler {
```
boubaker: Since it's almost the same behavior than `PortalRequestHandler`, would it better to handle standalone pages in `PortalRequestHandler` rather than duplicating the code ? WDYT ?

rdenarie: When I tried to use the same handler, I didnt manage to make it work correctly. 

In fact, If I use only PortalResquestHandler, when I access to /portal/login this code 
`String requestSiteName = controllerContext.getParameter(REQUEST_SITE_NAME);
String requestSiteType = controllerContext.getParameter(REQUEST_SITE_TYPE);`

set requestSiteName to "login", and requestSiteType to "portal"
Then, it search a portal named "login", which not exists.

By having firstly PageRequestHandler, which search the page "login" in "global" portal, AND if not found we call PortalRequestHandler, we ensure to find the page before.

What I can try is to make PageRequestHandler extension of PortaltRequestHandler, to only override modified method, but I think we cannot make all the stuff in one handler

boubaker: You can make a list of extensible `paths`, which will make the handler handles the page as it's part of `global` site.
If the request URI is one of those paths, you will be able to force the Site Type and Name as made in `PageRequestHandler`. This list of paths can be provided in controller.xml (as param) or using plugin definitions, which makes the list easily extensible by addon (Could redefine the Download Page of documents by example to be an editable page rather than a JSP).

boubaker: We may even imagine something more dynamic. All page requests not having a valid Site will be considered as a request to display a global site navigation, WDYT ?

rdenarie: As I understand, you mean to list in controller.xml all pages which are in global site which should be accessible on /portal/xxx
Is it right  ?

The current solution offers the advantage to work for all pages : 
when accessing to /portal/pageName, if the page is present in global navigation, it works, without having to declare it in  addon. 
Example page : /portal/download-document for public document download, on exoplatform side

rdenarie: > We may even imagine something more dynamic. All page requests not having a valid Site will be considered as a request to display a global site navigation, WDYT ?

how do you define "valid site" ? 
Should I call processRequest(context, app); firstly, and if nothing found, recall it with "global" sitename ?

boubaker: nope, before `processRequest` (see attachment)

image

component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
rdenarie: done

component/web/security/src/main/java/org/exoplatform/web/login/LoginHandler.java
rdenarie: done

component/web/security/src/main/java/org/exoplatform/web/login/onboarding/OnboardingHandler.java
rdenarie: done

webui/src/main/java/org/exoplatform/portal/application/PageRequestHandler.java
rdenarie: Done.
I had to add a specific boolean globalPageAccess

This is used to prevent the access a page through a url like /portal/global/pageName to keep current behaviour

boubaker: If we let accessing global pages, from now on using `/portal/{nav_name}`, we should allow to use `/portal/global/{nav_name}`. WDYT ?

rdenarie: No pb for me.

@srenault-meeds are you ok with this ?

srenault-meeds: ok for me too

rdenarie: Done

boubaker (approved):

PR #1173, opened 2025-09-12
fix: Unable to reload page configuration when addons add another page in the portal - MEED-9696

rdenarie:
Before this fix, when pages are configured in at least two pages/navigation.xml, only the first one is reload when using Reload Page configuration feature. This is because, when searching PortalConfig, we treat only the first one. This commit updates it to iterate on each existing portal configs, so that, all pages are correctly reloaded

Resolves meeds-io/meeds#3630


rdenarie:
@boubaker If I remember well, you worked on restore configuration feature.

Can you take a quick look on this fix, and see if you have remark or objections ? 

We discover it during development of MIPs#203, so I fix it in the process.

boubaker (approved):

PR #1185, opened 2025-10-16
fix : Impossible to require forgot password - MEED-9855

rdenarie:
Before this fix, the forgot password form is not accessible, the link to access is only /portal In addition the link in login form email is not complete

This is because the route for forgot-password have been removed from controller.xml because no more used, and it was used in PasswordRecoveryService This commit ensure to build the url without the route. It also apply for external-registration and on-boarding routes.

Resolves meeds-io/meeds#3747


azayati (changes requested):
merge should be done after changing poms

azayati (approved):

PR #1216, opened 2026-01-15
fix: Redirect to login page when requesting anonymously a restricted global page - MEED-9940

ahamdi:
When requesting a global page anonymously, the first accessible page is returned and we have some issues in the display (Login page with some alerts from the inaccessible page)
The fix : 
 - Anonymous user : if the page is not found or restricted, the request is redirected to the login page
 - Logged-in user : if the page is not found , a Page not found page is displayed. Otherwise, we preserve the current behavior of user redirections.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/config/UserPortalConfigService.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/config/UserPortalConfigService.java
@@ -418,7 +418,7 @@ public class UserPortalConfigService implements Startable {
       do {
         if (isAccessiblePageNoDraft(targetUserNode, username)) {
           return targetUserNode;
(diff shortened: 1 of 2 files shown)


ahamdi:
@boubaker I would like to have your feedback on this fix ? Thanks

boubaker:
> @boubaker I would like to have your feedback on this fix ? Thanks

I think that I can't answer to the question without studying the impact on existing flows and use cases. But I think we need to understand what are the use cases and flows to preserve and which ones to update. I let you schedule a meeting in order to collaborate on this, WDYT ?

ahamdi:
@boubaker as discussed with @srenault-meeds @Julien-Dubois-eXo , the new PR will implement the requested behavior
Could you take a look and review . Thanks

boubaker (approved):
Thanks @ahamdi

PR #1243, opened 2026-03-25
fix: Handle mapped LDAP attributes missing for users - MEED-10199 - Meeds-io/meeds#3736

hakermi:
Currently, when a user has a mapped attribute defined in the LDAP configuration but the attribute is not set in their LDAP entry, it is simply ignored. This causes Social profiles to retain stale values even after the attribute is removed from LDAP.

This change updates the attribute mapping logic to:

- Iterate over all mapped attributes.
- If a mapped attribute exists in configuration but is not present for the user in LDAP, set its value to null in the IdentityObjectAttribute.
- Preserve existing logic for attributes that are present in LDAP.

This ensures that during downstream synchronization, attributes removed from LDAP will also be cleared from the  Social profile, preventing stale data.

### DIFF
--- a/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
+++ b/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
@@ -653,10 +653,9 @@ public class ExoLDAPIdentityStoreImpl extends LDAPIdentityStoreImpl {
               identityObjectAttribute.addValue(value);
             }
           }
-        } else{
+        } else {
           // mapped attribute exists in configuration but not set for user in LDAP
           identityObjectAttribute.addValue(null);
-          log.info("Mapped attribute '" + attrName + "' exists in configuration but is not set for user: " + dn);
         }
         attrsMap.put(name, identityObjectAttribute);
       }
--- a/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
+++ b/component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
@@ -628,20 +628,17 @@ public class ExoLDAPIdentityStoreImpl extends LDAPIdentityStoreImpl {
       Name jndiName = new CompositeName().add(dn);
       Attributes attrs = ldapContext.getAttributes(jndiName);
 
-      for (Iterator<String> iterator = mappedNames.iterator(); iterator.hasNext();)
-      {
-        String name = iterator.next();
+      for (String name : mappedNames) {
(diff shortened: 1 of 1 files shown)


component/identity/src/main/java/org/picketlink/idm/impl/store/ldap/ExoLDAPIdentityStoreImpl.java
```diff
@@ -650,16 +647,18 @@ public Map<String, IdentityObjectAttribute> getAttributes(IdentityStoreInvocatio
...
               identityObjectAttribute.addValue(value);
             }
           }
-          attrsMap.put(name, identityObjectAttribute);
-        } else {
-          log.fine("No such attribute ('" + attrName + "') in entry: " + dn);
+        } else{
+          // mapped attribute exists in configuration but not set for user in LDAP
+          identityObjectAttribute.addValue(null);
+          log.info("Mapped attribute '" + attrName + "' exists in configuration but is not set for user: " + dn);
```
boubaker: No need to inform on such behavior in log each time a user is mapped from LDAP

```suggestion
        } else {
          // mapped attribute exists in configuration but not set for user in LDAP
          identityObjectAttribute.addValue(null);
```

boubaker:
A small question about that, does this mean that all mapped LDAP attributes are in Read Only mode if it's set each time from LDAP ?

hakermi:
> A small question about that, does this mean that all mapped LDAP attributes are in Read Only mode if it's set each time from LDAP ?

Yes, looks like all mapped LDAP attributes are fully controlled by LDAP.
Any local changes will be overwritten on the next LDAP sync, or set to **null** if the attribute is absent in LDAP (this behavior was added in the above change). 
From a downstream perspective, they effectively behave as read-only.

boubaker:
> From a downstream perspective, they effectively behave as read-only.

Thanks, FYI @srenault-meeds 

Shouldn't we make this more explicit by making the **Property Setting** definition as `Readonly` when it's mapped from LDAP ?

boubaker:
(May be a different concern and issue, I let you @srenault-meeds decide about the `Readonly` behavior)

srenault-meeds:
> > From a downstream perspective, they effectively behave as read-only.
> 
> Thanks, FYI @srenault-meeds
> 
> Shouldn't we make this more explicit by making the **Property Setting** definition as `Readonly` when it's mapped from LDAP ?

Actually, customers manually update the configuration to make it not editable.

boubaker (approved):
ok for me

boubaker:
> Actually, customers manually update the configuration to make it not editable.

I propose to make an enhancement for this to make it automatic rather than making this point as a manual LDAP installation step, WDYT @srenault-meeds ?

srenault-meeds:
> > Actually, customers manually update the configuration to make it not editable.
> 
> I propose to make an enhancement for this to make it automatic rather than making this point as a manual LDAP installation step, WDYT @srenault-meeds ?

Yes indeed. I will suggest an enhancement issue. Thanks for the suggestion.

PR #1263, opened 2026-05-18
feat: List nested groups in groups page - MEED-10390 - Meeds-io/MIPs#237

sofyenne:
This change introduces a REST endpoints to lists nested groups based on the parent group identifier.

### DIFF
--- a/component/portal/src/main/java/org/exoplatform/portal/rest/GroupRestResourcesV1.java
+++ b/component/portal/src/main/java/org/exoplatform/portal/rest/GroupRestResourcesV1.java
@@ -38,6 +38,7 @@ import io.swagger.v3.oas.annotations.parameters.RequestBody;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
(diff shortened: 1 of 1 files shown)


component/portal/src/main/java/org/exoplatform/portal/rest/GroupRestResourcesV1.java
```diff
@@ -692,6 +699,64 @@ public Response getGroupsTreeMembers(@Context
...
+                                                return organizationService.getGroupHandler().findGroupById(id);
+                                              } catch (Exception e) {
+                                                return null;
+                                              }
+                                            })
+                                            .filter(Objects::nonNull)
+                                            .toList();
+    int totalSize = groups.size();
+    List<Group> paginatedGroups = groups.stream().skip(offset).limit(limit).toList();
+    CollectionEntity<Group> result = new CollectionEntity<>(paginatedGroups, offset, limit, returnSize ? totalSize : 0);
```
boubaker: Return the same Object type than other REST methods: `GroupRestEntity`

sofyenne: done thanks!

boubaker (approved):

PR #1284, opened 2026-07-07
feat: Improve the loading of users by group - MEED-10462 - Meeds-io/MIPs#237

sofyenne:


component/identity/src/main/java/org/exoplatform/services/organization/idm/UserDAOImpl.java
```diff
@@ -795,6 +795,27 @@ public int getSize() throws Exception {
     return new IDMUserListAccess(qb, 20, false, countPaginatedUsers(), userStatus);
   }
 
+  @Override
+  public List<String> findUserNamesByGroupId(String groupId) {
```
boubaker, line 799: Shouldn't we add pagination & count ?

